Issue #2, opened 2023-07-11
Ensure the include/meson.build is adequate


WillPower3309:
I may have over trimmed the `include/meson.build` file. We should ensure that it is appropriately set up

PR #4, opened 2023-07-11
Sucessfully compile libscenefx and tinywl

Shinyzenith:
Closes: https://github.com/wlrfx/scenefx/issues/2


Shinyzenith:
Everything builds and compiles fine when I disable `werror=true` in meson.build. I don't know why that seems to be an issue for me but I hope others can maybe shed some light on it.

Issue #8, opened 2023-07-15
Pass eyecandy data to the scene API


WillPower3309:
More or less a tracking issue of https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3393. We should try to solve this upsteam, and then on our end, pass the decoration data the same way that opacity is passed.

ErikReider:
Seems like it's already been done (the opacity bit) 😅

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4145

WillPower3309:
Haha I had to laugh at the dialogue, good on you for messaging about it

WillPower3309:
Could do something similar but with `wlr_scene_buffer_set_effects`? Then everything else could be identical, compositors would only need to call that function

ErikReider:
We should also be able to pass global args to the fx_renderer. One example would be a blur toggle

Shinyzenith:
> We should also be able to pass global args to the fx_renderer. One example would be a blur toggle

Has this landed in tinywl main branch?

ErikReider:
> > We should also be able to pass global args to the fx_renderer. One example would be a blur toggle
> 
> Has this landed in tinywl main branch?

No. Just a reminder that we have to add this functionality :)

WillPower3309:
> We should also be able to pass global args to the fx_renderer. One example would be a blur toggle

I'm actually not sure about this. In the SwayFX implementation, fx_renderer has no knowledge of config values, and I'm wondering if we should keep it like that for scenefx. It would simplify things a bit if we just blurred windows when specifically told, and had the compositor handle always telling us to blur if a global setting is set

ErikReider:
> > We should also be able to pass global args to the fx_renderer. One example would be a blur toggle
> 
> I'm actually not sure about this. In the SwayFX implementation, fx_renderer has no knowledge of config values, and I'm wondering if we should keep it like that for scenefx. It would simplify things a bit if we just blurred windows when specifically told, and had the compositor handle always telling us to blur if a global setting is set

Aaah, very true! A global would be kinda redundant if it already knows that it has to blur. How are we gonna tackle this? Use the method that's used in the `deco_data_temp` branch or something else?

WillPower3309:
Let's bring the deco_data_temp branch into a PR and we can iterate from there

PR #10, opened 2023-07-19
Add corner radius parameter

ErikReider:
Fixes: #8


WillPower3309:
Will review this tomorrow! Great stuff Erik :)

include/types/decoration_data.h
```diff
@@ -0,0 +1,31 @@
...
+	int corner_radius;
+	// TODO: Change name from dim to something like "overlay" (maybe replace
+	// with it's own struct?)
+	float dim;
+	float *dim_color;
+	// TODO: Change to `enum corner_location` instead?
+	bool has_titlebar;
+	// TODO: Remove blur bool and add multiple fields (maybe even it's own
+	// struct?)
+	bool blur;
```
ErikReider: We should remove these unused fields

WillPower3309: Yeah, for now let's just implement a `wlr_scene_buffer_set_corner_radius` function. From there, we can expand out as needed as more effects are added

tinywl/tinywl.c
```diff
@@ -552,6 +596,45 @@ static void server_cursor_frame(struct wl_listener *listener, void *data) {
...
+		if (deco_data &&
+				xdg_surface &&
+				xdg_surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL) {
+			// TODO: Be able to set whole decoration_data instead of calling
+			// each individually?
+			wlr_scene_buffer_set_opacity(buffer, deco_data->alpha);
+
+			if (!wlr_surface_is_subsurface(xdg_surface->surface)) {
+				wlr_scene_buffer_set_corner_radius(buffer, deco_data->corner_radius);
+			}
```
ErikReider: What's your opinion on this? Maybe add a `wlr_scene_buffer_set_decoration_data` while also keeping the other functions?

WillPower3309 (changes requested):
For now, let's completely remove the decoration data struct. We can just set corner radius the same way opacity is set. My thoughts are that we may want to revisit how we break down the structure of the eye candy options as we add more to scenefx. We can always add deco data later if we decide we need it, it was more needed to simplify the iterators in swayfx

tinywl/tinywl.c
```diff
@@ -90,6 +103,10 @@ struct tinywl_view {
 	struct wl_listener request_maximize;
 	struct wl_listener request_fullscreen;
 	int x, y;
+
+	struct wlr_addon addon;
+
+	struct tinywl_decoration_data deco_data;
```
WillPower3309: Let's have the corner radius and alpha be in the tinywl view struct, no need for a decoration data struct

tinywl/tinywl.c
```diff
@@ -102,6 +119,44 @@ struct tinywl_keyboard {
...
+static struct tinywl_view *tinywl_view_addon_get(struct wlr_addon_set *addons,
+		const void * owner) {
+	struct wlr_addon *addon =
+		wlr_addon_find(addons, owner, &tinywl_view_addon_impl);
+	if (addon == NULL) {
+		return NULL;
+	}
+	struct tinywl_view *view = wl_container_of(addon, view, addon);
+	return view;
+}
```
WillPower3309, line 147: QQ: where is this being added from?

WillPower3309 (changes requested):
One question and one tiny comment

tinywl/tinywl.c
ErikReider: Just some logic to add the view as a addon to the scene_tree so that we're able to get our decoration data. The idea comes from the Sway scene PR :)

WillPower3309 (approved):

PR #12, opened 2023-07-20
feat: tinywl: Simple static borders

Shinyzenith:


lidgnulinux:
Nice feature, unfortunately the borders are formed inside the window, it will cover window surface. Would be nice if the borders are formed outside the window.

Shinyzenith:
Hi! This patch was just for testing purposes I think we can go ahead and close this.

Issue #13, opened 2023-07-27
Add box shadows


WillPower3309:
We should add the box shadows from upstream SwayFX to scenefx

WillPower3309:
We should have a discussion about how we want this feature to fit into the scene API design and how it should interact with borders. In my mind we have two options:
1. Have a wlr_box parameter that users will use to specify the area they'd like the shadow to be drawn around (that way they can account for borders)
2. Have the wlr_box be determined based on the scene buffer, and tied to borders (which would also have to be tied to the scene buffer). This is my preferred option, as while it does force the user to create borders through us, this will pretty much be required to support rounded borders anyway (also, less configuration/ less to worry about for the user)

ErikReider:
> I agree with your second option. So we'll have to add border support first before shadows

In the #16 PR we grab a stencil the size of the surface geometry, we could reuse that stencil for drawing the single, large border.

IMO, we should provide a border horizontal_size, vertical_size, a x and y offset, and of course some field for providing color(s)

WillPower3309:
> I agree with your second option. So we'll have to add border support first before shadows

Not necessarily, as shadows need to work without borders too :) we'll just add support for making them work w borders as we merge borders

WillPower3309:
> > I agree with your second option. So we'll have to add border support first before shadows
> 
> In the #16 PR we grab a stencil the size of the surface geometry, we could reuse that stencil for drawing the single, large border.
> 
> IMO, we should provide a border horizontal_size, vertical_size, a x and y offset, and of course some field for providing color(s). Maybe also a border radius field too but who would use a different one from the surface one though?
> 
> Edit: We should actually check how eg the Sway scene PR adds borders

I'd suspect they use the scene rect rendering function, but we may not be able to reuse the stencil as we'd need the stencil for the shadow to encompass the border region too unfortunately, unless we can resize it.

Let's keep the same radius as the surface too

PR #16, opened 2023-07-27
Add box shadows

ErikReider:
Fixes: #13


WillPower3309:
Let's worry about border compensation as we merge borders, shouldn't be too hard, and we need to support shadows without borders anyway

types/scene/wlr_scene.c
```diff
@@ -1148,6 +1238,38 @@ static void scene_node_render(struct fx_renderer *fx_renderer, struct wlr_scene_
 		wlr_matrix_project_box(matrix, &dst_box, transform, 0.0,
 			output->transform_matrix);
 
+		// Some surfaces (mostyly GTK 4) decorate their windows with shadows
+		// which extends the node size past the actual window size. This gets
+		// the actual surface geometry, mostyly ignoring CSD decorations
+		// but only if we need to.
+		if (scene_buffer->corner_radius != 0 ||
```
WillPower3309, line 1233: Would the corner radius be important to this check?

types/scene/wlr_scene.c
```diff
@@ -1148,6 +1238,38 @@ static void scene_node_render(struct fx_renderer *fx_renderer, struct wlr_scene_
...
+				shadow_data_is_enabled(&scene_buffer->shadow_data)) {
```
WillPower3309: Let's use a more descriptive name, like `scene_buffer_has_shadow()`

types/scene/wlr_scene.c
```diff
@@ -1097,6 +1136,57 @@ static void render_texture(struct fx_renderer *fx_renderer, struct wlr_output *o
...
+	inner_box.x += corner_radius * 0.5;
+	inner_box.y += corner_radius * 0.5;
+	inner_box.width -= corner_radius;
+	inner_box.height -= corner_radius;
+	pixman_region32_t inner_region = create_damage(inner_box, output_damage);
+	pixman_region32_subtract(&damage, &damage, &inner_region);
+
+	if (!pixman_region32_not_empty(&damage)) {
+		goto damage_finish;
+	}
```
WillPower3309: lets take the damage from `scene_node_render` as a parameter in this function (like we do in the other rendering functions). We can get rid of the `create_damage` function and then scale / subtract from the damage parameter as needed here.

types/scene/wlr_scene.c
```diff
@@ -240,6 +254,7 @@ static void scene_node_opaque_region(struct wlr_scene_node *node, int x, int y,
 			return;
 		}
 
+		/* Buffer is translucent*/
```
WillPower3309: nit: use `//`

types/scene/wlr_scene.c
```diff
@@ -398,6 +413,15 @@ static bool scene_node_update_iterator(struct wlr_scene_node *node,
 		pixman_region32_fini(&opaque);
 	}
 
+	/* Expand the nodes visible region by the shadow size */
```
WillPower3309: nit: use `//`

types/scene/wlr_scene.c
WillPower3309: The expansion may be already done for us in the `scene_node_update_iterator` changes you made. We may be able to then also remove the `#include <pixman.h>`

types/scene/wlr_scene.c
```diff
@@ -1148,6 +1238,38 @@ static void scene_node_render(struct fx_renderer *fx_renderer, struct wlr_scene_
 		wlr_matrix_project_box(matrix, &dst_box, transform, 0.0,
 			output->transform_matrix);
 
+		// Some surfaces (mostly GTK 4) decorate their windows with shadows
+		// which extends the node size past the actual window size. This gets
+		// the actual surface geometry, mostyly ignoring CSD decorations
```
WillPower3309: typo: `mostyly`

types/scene/wlr_scene.c
```diff
@@ -1,9 +1,12 @@
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <pixman.h>
 #include <stdlib.h>
 #include <string.h>
 #include <wlr/backend.h>
 #include <wlr/render/gles2.h>
+#include <wlr/types/wlr_xdg_shell.h>
+#include <wlr/types/wlr_layer_shell_v1.h>
```
WillPower3309: Q: where do we use structures / functions from this header?

WillPower3309 (changes requested):
Q: why the increased complexity over the initial stencil buffer implementation in the [box_shadow PR for swayfx](https://github.com/WillPower3309/swayfx/pull/64)?

ErikReider:
> Q: why the increased complexity over the initial stencil buffer implementation in the https://github.com/WillPower3309/swayfx/pull/64?

Good point! I'll simplify the stencil init function :)

types/scene/wlr_scene.c
ErikReider: I just renamed the function name. Not sure if you wanted it to be in `wlr_scene.c` or in `shadow_data.c` though

types/scene/wlr_scene.c
ErikReider: Otherwise the corners wouldn't be clipped if the comp would only specify to change the corner_radius due to the GTK CSD shadows extending the surface rect:

image

The white area being the extra space

include/render/fx_renderer/fx_stencilbuffer.h
```diff
@@ -0,0 +1,20 @@
...
+	GLuint rb;
+	int width;
+	int height;
+};
+
+struct fx_stencilbuffer fx_stencilbuffer_create(void);
+
+void fx_stencilbuffer_init(struct fx_stencilbuffer *stencil_buffer, int width, int height);
+
+void fx_stencilbuffer_release(struct fx_stencilbuffer *stencil_buffer);
```
WillPower3309, line 18: looks like we don't use this function, should we call it in fx_renderer, or is it better to remove it?

WillPower3309 (commented):
One last comment

include/render/fx_renderer/fx_stencilbuffer.h
ErikReider: Done :)

Forgot to call it on display destruction...

WillPower3309 (approved):
LGTM

PR #23, opened 2023-10-19
feat: add borders

WillPower3309:
TODO:

+ [x] border setting
+ [x] add border width to corner radius
+ [x] border stencil
+ [x] proper damage tracking
+ [ ] support scaled and rotated outputs
+ [ ] decrease buffer size by border size (otherwise the borders can mess up gaps because they increase the visual size of the surface)
+ [x] remove slight rounding with corner radius of 0

### DIFF
--- a/render/fx_renderer/fx_renderer.c
+++ b/render/fx_renderer/fx_renderer.c
@@ -110,9 +110,9 @@ static bool link_quad_rounded_program(struct quad_rounded_shader *shader) {
 	shader->proj = glGetUniformLocation(prog, "proj");
 	shader->color = glGetUniformLocation(prog, "color");
 	shader->pos_attrib = glGetAttribLocation(prog, "pos");
-	shader->half_size = glGetAttribLocation(prog, "half_size");
-	shader->position = glGetAttribLocation(prog, "position");
-	shader->radius = glGetAttribLocation(prog, "radius");
+	shader->half_size = glGetUniformLocation(prog, "half_size");
+	shader->position = glGetUniformLocation(prog, "position");
+	shader->radius = glGetUniformLocation(prog, "radius");
 
 	return true;
 }
@@ -529,7 +529,7 @@ void fx_render_rounded_rect(struct fx_renderer *renderer, const struct wlr_box *
 		return;
 	}
 	assert(box->width > 0 && box->height > 0);
-	printf("corner_radius: %d\n", corner_radius);
+
 	if (corner_radius == 0) {
 		fx_render_rect(renderer, box, color, projection);
 		return;
--- a/render/fx_renderer/fx_renderer.c
+++ b/render/fx_renderer/fx_renderer.c
@@ -523,54 +523,6 @@ void fx_render_rect(struct fx_renderer *renderer, const struct wlr_box *box,
 	glDisableVertexAttribArray(shader.pos_attrib);
 }
 
-void fx_render_rounded_rect(struct fx_renderer *renderer, const struct wlr_box *box,
-		const float color[static 4], const float projection[static 9], int corner_radius) {
-	if (box->width == 0 || box->height == 0) {
-		return;
-	}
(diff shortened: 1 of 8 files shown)


WillPower3309:
will fix `meson.build` next commit

WillPower3309:
Need to fix a weird bug, corners of the borders don't round, and the stencil cuts out the window surface texture as well as the border area

render/fx_renderer/fx_renderer.c
```diff
@@ -539,6 +562,63 @@ static void fx_render_stencil_mask(struct fx_renderer *renderer,
...
+	// wlr_matrix_multiply(gl_matrix, flip_180, gl_matrix);
+
+	wlr_matrix_transpose(gl_matrix, gl_matrix);
+
+	if (corner_radius) {
+		// Init stencil work
+		fx_renderer_stencil_mask_init();
+		// Draw the rounded rect as a mask
+		fx_render_stencil_mask(renderer, stencil_box, matrix, corner_radius);
+		fx_renderer_stencil_mask_close(false);
```
ErikReider, line 593: Don't forget to call the stencil fini function when done

types/scene/wlr_scene.c
```diff
@@ -401,12 +401,13 @@ static bool scene_node_update_iterator(struct wlr_scene_node *node,
...
 	}
 
-	// Expand the nodes visible region by the shadow size
+	// Expand the nodes visible region by the border & shadow size
 	if (node->type == WLR_SCENE_NODE_BUFFER) {
 		struct wlr_scene_buffer *buffer = wlr_scene_buffer_from_node(node);
-		struct shadow_data *data = &buffer->shadow_data;
-		if (scene_buffer_has_shadow(data)) {
-			wlr_region_expand(&node->visible, &node->visible, data->blur_sigma);
+		wlr_region_expand(&node->visible, &node->visible, buffer->border_size);
```
ErikReider, line 407: Maybe check if the size is larger than 0?

types/scene/wlr_scene.c
```diff
@@ -1124,10 +1139,54 @@ static void render_texture(struct fx_renderer *fx_renderer, struct wlr_output *o
 	}
 }
 
+static void render_border(struct fx_renderer *fx_renderer, struct wlr_output *output,
+		pixman_region32_t *surface_damage, const struct wlr_box *surface_box,
+		int size, float color[static 4], int corner_radius) {
```
ErikReider: What about passing the scene_buffer pointer instead of these three parameters?

ErikReider (changes requested):

ErikReider:
What about moving this into a sister-node to our `wlr_scene_buffer` as a new node type? We could provide shadow control, border x/y offset, width, height, etc without messing with the `wlr_scene_buffer` struct as intended by the wlroots devs:

image

But then again, we could just add those variables to the `wlr_scene_buffer` struct and get the same result

WillPower3309:
> What about moving this into a sister-node to our `wlr_scene_buffer` as a new node type? We could provide shadow control, border x/y offset, width, height, etc without messing with the `wlr_scene_buffer` struct as intended by the wlroots devs:
> 
> image
> 
> This would allow the comp to add multiple borders as discussed earlier.
> 
> But then again, we could just add those variables to the `wlr_scene_buffer` struct and get the same result

Thanks for closing out this couple of items I had to do, I agree. I was actually going to suggest something similar, partly since shadows also depend on border values

WillPower3309:
Just need to fix a corner issue and close out the scaling / rotation and this is done!

WillPower3309:
I'll try to close out the last bit tomorrow

WillPower3309:
Still a bit of corner cropping, almost seems like the stencil is also applying to the surface texture

WillPower3309:
Might rename the function to render_rect_around_rect, to make it less specific to borders, as compositors should be able to use it for titlebars too

Issue #26, opened 2024-01-06
Example tinywl : missing headers ?


lidgnulinux:
Hello, when I built the example tinywl, it's failed. It seems that there are missing headers. 

```
#include <types/fx/shadow_data.h>
......
#include "types/wlr_scene.h"

#include "render/fx_renderer/fx_renderer.h"
```

Is it intentional ? Thanks.

ErikReider:
At the moment, you'll probably have to use scenefx as a subproject

ahmadraniri:
Facing the same issue here, @ErikReider subproject of what actually ? The tinywl ?

ErikReider:
What are you trying to build? The tinywl in the repo or a separated tinywl in a different repo?

ErikReider:
Oh now I know what you mean, mb. Fixing....

lidgnulinux:
Hello, Quick trial on the pull request #27 and I still face the missing headers complain.

lidgnulinux:
Hello, Quick trial on your PR and I'm still facing the missing headers complain.

```
/home/lidg/.local/include/scenefx/fx_renderer/fx_renderer.h:15:10: fatal error: render/pass.h: No such file or directory
   15 | #include "render/pass.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:16: tinywl] Error
```

ErikReider:
> Hello, Quick trial on your PR and I'm still facing the missing headers complain.
> 
> ```
> /home/lidg/.local/include/scenefx/fx_renderer/fx_renderer.h:15:10: fatal error: render/pass.h: No such file or directory
>    15 | #include "render/pass.h"
>       |          ^~~~~~~~~~~~~~~
> compilation terminated.
> make: *** [Makefile:16: tinywl] Error
> ```

You should compile everything with meson, not the makefile

ahmadraniri:
Oh my bad.

lidgnulinux:
> You should compile everything with meson, not the makefile


I talked about the example tinywl and not the scenefx, using meson give me this : 

```
Version: 1.3.0
Source dir: /home/lidg/cmp/scenefx/tinywl
Build dir: /home/lidg/cmp/scenefx/tinywl/build
Build type: native build

ERROR: Not the project root: first statement must be a call to project()

Did you mean to run meson from the directory: "/home/lidg/cmp/scenefx"?

[...]

PR #27, opened 2024-01-06
Meson fixes

ErikReider:
Fixes: #26 

Todo:
- [x] Add back xkbcommon dep
- [x] Move some headers into `scenefx` dir
	- [x] Shadow struct
	- [x] custom wlr_scene struct
	- [x] All of the fx_renderer/pass headers?
- [x] Fix not being able to build statically due to conflicting functions


ErikReider:
I've tested it and everything works as expected, even installed it as a library on my system :)

The only question mark is how we're going to expose the fx_renderer. Everything in the renderer is exposed atm which might not be ideal. Could be useful for SwayFX though?

Any input?

WillPower3309:
> I've tested it and everything works as expected, even installed it as a library on my system :)
> 
> The only question mark is how we're going to expose the fx_renderer. Everything in the renderer is exposed atm which might not be ideal. Could be useful for SwayFX though?
> 
> Any input?

Discussed on discord but mentioning here too for anyone following the issue on GitHub, for now we will be exposing the bare minimum functions for SwayFX to use. We can always expose more functions but can't take them away (without annoying our users) so we're starting small.

ErikReider:
Should be ready now

ErikReider:
I wonder if we could remove our own declaration of `struct wlr_scene_*` and just rename the only necessary function names to fx_*? I'll investigate...

WillPower3309 (approved):

lidgnulinux:
For the example tinywl, I think the Makefile need little additional : 
```
$(shell pkg-config --cflags --libs scenefx) 
```

Issue #14, opened 2023-07-27
Add blur


WillPower3309:
We should add the blur from upstream SwayFX to scenefx

WillPower3309:
Let's wait to complete this task until https://github.com/WillPower3309/swayfx/pull/181 is done

Shinyzenith:
Related: https://github.com/WillPower3309/swayfx/pull/219

https://github.com/WillPower3309/swayfx/pull/181 was superceeded by 219

PR #22, opened 2023-10-16
Blur

ErikReider:
Fixes: #14 

Note: The blur isn't correctly aligned but should™ be easily fixed

TODO:
- [x] Add global config options (per surface blur won't work due to the damage calculations)
- [x] Extend the damage (fixing artifacts)
- [ ] Blit the old region to remove artifacts
- [x] Fix blur not rendering correctly
- [x] Add tinywl example
- [ ] Refactor/simplify `fx_texture` code?
- [ ] Add "optimized" blur

PR #30, opened 2024-01-28
Add blur

ErikReider:
Fixes: #14 
Supersedes: #22 

TODO:
- [x] Add global config options (per surface blur won't work due to the damage calculations)
- [x] Extend the damage (fixing artifacts)
- [ ] Add tinywl example
- [ ] Add "optimized" blur

### DIFF
--- a/include/render/fx_renderer/fx_renderer.h
+++ b/include/render/fx_renderer/fx_renderer.h
@@ -48,6 +48,10 @@ struct fx_framebuffer {
 	struct wlr_addon addon;
 };
 
+/** Should only be used with custom fbs */
+void fx_framebuffer_get_or_create_bufferless(struct fx_renderer *fx_renderer,
+		struct wlr_output *output, struct fx_framebuffer **fx_buffer);
+
 struct fx_framebuffer *fx_framebuffer_get_or_create(struct fx_renderer *renderer,
 		struct wlr_buffer *wlr_buffer);
 
@@ -141,20 +145,28 @@ struct fx_renderer {
 		PFNGLGETINTEGER64VEXTPROC glGetInteger64vEXT;
 	} procs;
 
-	struct {
-		struct quad_shader quad;
-		struct tex_shader tex_rgba;
-		struct tex_shader tex_rgbx;
-		struct tex_shader tex_ext;
-		struct box_shadow_shader box_shadow;
-		struct stencil_mask_shader stencil_mask;
-	} shaders;
+	struct shaders shaders;
 
 	struct wl_list buffers; // fx_framebuffer.link
 	struct wl_list textures; // fx_texture.link
 
 	struct fx_framebuffer *current_buffer;
 	uint32_t viewport_width, viewport_height;
+
+	// Contains the blurred background for tiled windows
+	struct fx_framebuffer *blur_buffer;
+	// Contains the original pixels to draw over the areas where artifact are visible
+	struct fx_framebuffer *blur_saved_pixels_buffer;
+	// Blur swaps between the two effects buffers everytime it scales the image
+	// Buffer used for effects
+	struct fx_framebuffer *effects_buffer;
+	// Swap buffer used for effects
+	struct fx_framebuffer *effects_buffer_swapped;
+
+	// The region where there's blur
+	pixman_region32_t blur_padding_region;
+
+	bool blur_buffer_dirty;
 };
 
 bool wlr_renderer_is_fx(struct wlr_renderer *wlr_renderer);
--- a/include/render/fx_renderer/shaders.h
+++ b/include/render/fx_renderer/shaders.h
@@ -37,6 +37,7 @@ struct tex_shader {
 	GLint size;
 	GLint position;
 	GLint radius;
+	GLint discard_transparent;
 };
 
 struct stencil_mask_shader {
@@ -58,6 +59,27 @@ struct box_shadow_shader {
 	GLint size;
 	GLint blur_sigma;
 	GLint corner_radius;
+};
+
+struct blur_shader {
+	GLuint program;
+	GLint proj;
+	GLint tex_proj;
+	GLint tex;
+	GLint pos_attrib;
+	GLint radius;
+	GLint halfpixel;
+};
+
+struct shaders {
+	struct quad_shader quad;
+	struct tex_shader tex_rgba;
+	struct tex_shader tex_rgbx;
+	struct tex_shader tex_ext;
+	struct box_shadow_shader box_shadow;
+	struct stencil_mask_shader stencil_mask;
+	struct blur_shader blur1;
+	struct blur_shader blur2;
 };
 
 bool link_shaders(struct fx_renderer *renderer);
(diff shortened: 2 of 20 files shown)


include/render/fx_renderer/fx_renderer.h
```diff
@@ -141,20 +145,28 @@ struct fx_renderer {
...
 
-	struct {
-		struct quad_shader quad;
-		struct tex_shader tex_rgba;
-		struct tex_shader tex_rgbx;
-		struct tex_shader tex_ext;
-		struct box_shadow_shader box_shadow;
-		struct stencil_mask_shader stencil_mask;
-	} shaders;
+	struct shaders shaders;
```
WillPower3309: I think we actually want to keep this initialization as is here, as the renderer is the only place we use the `shaders` struct

include/render/pass.h
```diff
@@ -41,6 +43,16 @@ struct fx_render_rect_options {
 struct fx_render_rect_options fx_render_rect_options_default(
 		const struct wlr_render_rect_options *base);
 
+struct fx_render_blur_options {
```
WillPower3309: Nit: I feel as though `options` indicated that these values would impact the appearance of the blur, we may want to think of another name for this struct

include/scenefx/fx_renderer/fx_renderer.h
```diff
@@ -2,6 +2,7 @@
 #define SCENEFX_FX_OPENGL_H
 
 #include <wlr/backend.h>
+#include <wlr/render/allocator.h>
```
WillPower3309: Nit: if it isn't needed in the header we should include this in fx_renderer.c

render/fx_renderer/fx_framebuffer.c
```diff
@@ -20,6 +20,37 @@ static const struct wlr_addon_interface buffer_addon_impl = {
 	.destroy = handle_buffer_destroy,
 };
 
+void fx_framebuffer_get_or_create_bufferless(struct fx_renderer *renderer,
```
WillPower3309: Q: why is this function have `bufferless` in its name? This question might be from a lack of understanding of the wlr allocator but I'm curious about the naming here, seems it does create a buffer?

types/scene/wlr_scene.c
```diff
@@ -1733,6 +1847,55 @@ static bool scene_entry_try_direct_scanout(struct render_list_entry *entry,
...
 
+struct blur_info {
+	bool has_blur;
+	bool has_optimized;
+};
+
+static struct blur_info workspace_get_blur_info(int list_len,
+		struct render_list_entry *list_data, struct wlr_scene_output *scene_output,
+		pixman_region32_t *blur_region) {
+	// TODO: Check if a check for "is current workspace" is also needed
```
WillPower3309: commenting as a reminder to determine this

WillPower3309 (commented):
Some small comments, can you explain at a high level how we handle blur ignore transparent in scene? Reviewing from my phone and that bit of logic is the only one I'm having a nuisance tracking

WillPower3309:
Thanks for you patience in waiting for a review!

include/render/fx_renderer/fx_renderer.h
ErikReider: So you'd like to also move the shader.c impl into the renderer as well?

My reason for doing this is to make the fx_renderer.c file a bit shorter, making it easier for the eyes :)

render/fx_renderer/fx_framebuffer.c
ErikReider: The reason is that our custom fx_framebuffers aren't requesting the swapchain for a new wlr_buffer, so it's “bufferless”. I agree that the naming isn't the best, any input on a new name?

Maybe `fx_framebuffer_get_or_create_custom`?

types/scene/wlr_scene.c
ErikReider: Oh yeah, this isn't needed and already checked here:

https://github.com/wlrfx/scenefx/blob/e2e6d495b61a039600e825ae6cd6d5542a9eda21/types/scene/wlr_scene.c#L1990-L1991

include/scenefx/fx_renderer/fx_renderer.h
ErikReider: Done

include/render/pass.h
ErikReider: Done

ErikReider:
> an you explain at a high level how we handle blur ignore transparent in scene?

It's actually pretty simple, the comp needs to call the `wlr_scene_buffer_set_backdrop_blur_ignore_transparent` function for the `wlr_scene_buffer` node:

https://github.com/ErikReider/fx-comp/blob/80a2c853a62dafecef149b00c6888f7622c2a643/src/desktop/xdg.c#L39

The actual wlr_scene part of it is the same as we do in SwayFX where we pass a bool to the blur function indicating whether we want to ignore the transparent regions or not. (Render the surface texture and stencil it, etc…)

ErikReider:
What we can do is to separate the renderer impl and the scene impl into separate PRs to make it easier for you to review. What do you think of that idea @WillPower3309 ?

WillPower3309:
> What we can do is to separate the renderer impl and the scene impl into separate PRs to make it easier for you to review. What do you think of that idea @WillPower3309 ?

If you don't mind that would be great!

ErikReider:
> > What we can do is to separate the renderer impl and the scene impl into separate PRs to make it easier for you to review. What do you think of that idea @WillPower3309 ?
> 
> If you don't mind that would be great!

Done :)

include/render/fx_renderer/fx_renderer.h
WillPower3309: nope, just the `shaders` struct. I don't think it needs to exist outside of the renderer as it is only referenced there. I do think that the specific shader structs (`quad_shader`, `tex_shader` etc) are great to have in a separate file though as you've done since we do have functions that set them up that can live in the separate `shaders.c` file, my reasoning is more that the `shaders` struct itself has no functions that modify it.

include/render/fx_renderer/shaders.h
```diff
@@ -60,6 +61,40 @@ struct box_shadow_shader {
...
+	struct quad_shader quad;
+	struct tex_shader tex_rgba;
+	struct tex_shader tex_rgbx;
+	struct tex_shader tex_ext;
+	struct box_shadow_shader box_shadow;
+	struct stencil_mask_shader stencil_mask;
+	struct blur_shader blur1;
+	struct blur_shader blur2;
+	struct blur_effects_shader blur_effects;
+};
```
WillPower3309: same as mentioned in `fx_renderer.h`

render/fx_renderer/fx_framebuffer.c
WillPower3309: I'd agree with `fx_framebuffer_get_or_create_custom`

include/render/fx_renderer/fx_renderer.h
WillPower3309: `link_shaders` could then be moved back to fx_renderer.c

WillPower3309 (commented):
Just a couple minor comments, otherwise LGTM

ErikReider:
Done! :)

WillPower3309 (approved):

PR #33, opened 2024-02-26
[WLR Scene] Add blur

ErikReider:
Depends on #30 

Also depends on #66 for the initial shadow_node impl which the blur_node should base off of. Could potentially also add a custom_node impl for both of those effect nodes?

Todo:
- [x] Figure out the best way for the compositor to know when to rerender the optimized blur
- [x] tinywl example for optimized blur
- [ ] ~Make a seperate blur_node instead of forcing it onto a scene_buffer~ (Will be difficult to separate the discard transparent logic, so we should wait with this)

Tesing:
- Run tinywl
- Or run [fx-comp](https://github.com/ErikReider/fx-comp/pull/1) (instructions in the README). NOTE: the default keyboard modifier is left alt


types/scene/wlr_scene.c
```diff
@@ -1311,8 +1445,63 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+
+			bool has_alpha = false;
+			if (scene_buffer->opacity < 1.0) {
+				has_alpha = true;
+				pixman_region32_union_rect(&opaque_region, &opaque_region, 0, 0, 0, 0);
+			} else {
+				struct wlr_scene_surface *scene_surface
+					= wlr_scene_surface_try_from_buffer(scene_buffer);
+				has_alpha = !scene_surface->surface->opaque;
+				pixman_region32_copy(&opaque_region, &scene_surface->surface->opaque_region);
```
ErikReider: Set has_alpha to true if there's no scene_surface?

ErikReider:
An optimized blur example in tinywl wouldn't really make it that tiny and simple anymore, so skipping that :)

WillPower3309:
> An optimized blur example in tinywl wouldn't really make it that tiny and simple anymore, so skipping that :)

Understandable, but do you mind running me through what's required on the comp side for getting this to work in scene? I'm following it along but a summary may be helpful

My thinking here is that one of the main goals of scene is for compositors to easily leverage our effects in very few LOC

ErikReider:
> do you mind running me through what's required on the comp side for getting this to work in scene?

So the optimized blur is a new scene node that should be inserted between regular scene trees, which would be between bottom and workspace in sways use-case. The comp is responsible for resizing the node to the desired size, which in most use-cases would be the monitor's size and coordinates.

The comp is also responsible for redrawing the optimized blur by marking it as “dirty”. Sway currently redraws the blur when any background or bottom layer surface commits.

So actually implementing it in a fully fledged comp doesn't require that many LOC, but adding an example in TinyWL would require adding extra non-blur related code (the extra layers and a good way of showing off the optimized blur)

ErikReider:
Optimized blur (x-ray) example in tinywl is now added. It's **very** simple but it works good enough to demonstrate what it does and how it works :)

image

*Notice how only the white rectangle gets blurred 😉*

ErikReider:
Might want to merge this after 0.2? Would give us some more time to think of a better impl, especially if we get a #60 impl merged

WillPower3309:
> Might want to merge this after 0.2? Would give us some more time to think of a better impl, especially if we get a #60 impl merged

Sounds good to me!

types/fx/blur_data.c
```diff
@@ -11,6 +11,10 @@ struct blur_data blur_data_get_default(void) {
 	};
 }
 
+bool scene_buffer_should_blur(bool backdrop_blur, struct blur_data *blur_data) {
+	return backdrop_blur && blur_data->radius > 0 && blur_data->num_passes > 0;
+}
```
WillPower3309: I think `is_blur_enabled` would be a more appropriate parameter name than `backdrop_blur`

include/scenefx/types/wlr_scene.h
```diff
@@ -113,6 +115,8 @@ struct wlr_scene {
 	enum wlr_scene_debug_damage_option debug_damage_option;
 	bool direct_scanout;
 	bool calculate_visibility;
+
+	struct blur_data blur_data;
```
WillPower3309, line 118: why do we set this at the scene level? `blur_data` contains `has_blur` and `has_optimized`, those params seem node specific to me

include/scenefx/types/wlr_scene.h
```diff
@@ -326,6 +339,9 @@ struct wlr_scene *wlr_scene_create(void);
 void wlr_scene_set_presentation(struct wlr_scene *scene,
 	struct wlr_presentation *presentation);
 
+/** Sets the global blur parameters */
+void wlr_scene_set_blur_data(struct wlr_scene *scene, struct blur_data blur_data);
```
WillPower3309, line 343: Q related to above, why do we need global blur params?

include/scenefx/types/wlr_scene.h
```diff
@@ -507,6 +547,36 @@ void wlr_scene_buffer_set_filter_mode(struct wlr_scene_buffer *scene_buffer,
...
+/**
+ * Tells the renderer to re-render the optimized blur for the specified output.
+ * A NULL output will re-render the optimized blur on every output.
+ * This is very expensive so should only be called when needed.
+ *
+ * An example use would be to call this when a "static" node changes, like a
+ * wallpaper.
+ */
+void wlr_scene_optimized_blur_mark_dirty(struct wlr_scene *scene,
+		struct wlr_output *output);
```
WillPower3309: It would be great if we could do this dynamically in scene and ditch this function. Considering that scene optimizes the rendering by ignoring regions below opaque regions of the nodes *above*, I'd bet we could flip the script on that damage tracking optimization to do this dynamically when nodes *below* are updated

types/scene/wlr_scene.c
```diff
@@ -280,6 +294,9 @@ static void scene_node_opaque_region(struct wlr_scene_node *node, int x, int y,
 			pixman_region32_translate(opaque, x, y);
 			return;
 		}
+	} else if (node->type == WLR_SCENE_NODE_OPTIMIZED_BLUR) {
+		// Always transparent
```
WillPower3309, line 301: I don't think this is true - we never want to render anything below the blur, since the blur itself has a constant alpha channel of 1. Although for the case of the marking blur as dirty we would want to consider the layer below, not sure what the correct solution looks like here, maybe we should check the `blur_buffer_dirty` value

types/scene/wlr_scene.c
```diff
@@ -1673,6 +1880,8 @@ static bool scene_node_invisible(struct wlr_scene_node *node) {
 		struct wlr_scene_shadow *shadow = wlr_scene_shadow_from_node(node);
 
 		return shadow->color[3] == 0.f;
+	} else if (node->type == WLR_SCENE_NODE_OPTIMIZED_BLUR) {
+		return false;
```
WillPower3309, line 1985: wouldn't this conditionally be true? the optimized blur is only visible behind nodes that have it set. I know this is tricky to implement though :S

We could keep it to false if you don't see this introducing any damage tracking perf hits. A higher level concern I have is that a big node that takes up the whole screen space may mess up damage tracking

types/scene/wlr_scene.c
```diff
@@ -1378,6 +1492,32 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+			transform = wlr_output_transform_compose(transform, data->transform);
+			struct fx_render_blur_pass_options blur_options = {
+				.tex_options = {
+					.base = {
+						.transform = transform,
+						.alpha = &opacity,
+						.blend_mode = WLR_RENDER_BLEND_MODE_NONE,
+						.dst_box = dst_box,
+					},
+					.clip_box = &dst_box,
```
WillPower3309, line 1625: I think we should handle this like we do other nodes, where we set `clip` to `&render_region`

types/scene/wlr_scene.c
```diff
@@ -1422,6 +1562,59 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+				has_alpha = true;
+				pixman_region32_union_rect(&opaque_region, &opaque_region, 0, 0, 0, 0);
+			} else if ((scene_surface = wlr_scene_surface_try_from_buffer(scene_buffer))) {
+				has_alpha = !scene_surface->surface->opaque;
+				pixman_region32_copy(&opaque_region, &scene_surface->surface->opaque_region);
+			} else {
+				has_alpha = pixman_region32_not_empty(&opaque);
+				scene_node_opaque_region(node, x, y, &opaque_region);
+				scale_output_damage(&opaque_region, data->scale);
+			}
```
WillPower3309: can't we use `scene_node_opaque_region()`?

types/fx/blur_data.c
WillPower3309: this is a nit, not a blocker by any means

types/scene/wlr_scene.c
WillPower3309: Seems we inherit this from the tex options - clip box will always equal dst box afaik, do we have a reason why we can't remove this field from the tex options?

types/scene/wlr_scene.c
ErikReider: So the optimized blur node doesn't actually render anything onto the scene tree. The only thing it does is fill the optimized blur texture with the blur when the client calls the mark dirty function.

So we will always want to render what's below the node

include/scenefx/types/wlr_scene.h
ErikReider: So the only thing that blur_data contains is the blur render parameters. We want each node to render the same strength of blur for consistent damage tracking. By setting it in the scene root, we make sure that the damage will be constant when blurred

include/scenefx/types/wlr_scene.h
```diff
@@ -152,6 +156,12 @@ struct wlr_scene_shadow {
 	float blur_sigma;
 };
 
+/** A scene-graph node telling SceneFX to render the optimized blur */
```
ErikReider, line 158: We should maybe write that it doesn't actually render anything onto the tree

include/scenefx/types/wlr_scene.h
ErikReider: What SwayFX does is that the optimized blur texture only gets updated when a bottom or background layer signals a commit (updates). I'm pretty sure that it would be impossible for us to check that while also giving the comp choices on how to render this (scene doesn't know where or even if layer surfaces exist).

types/scene/wlr_scene.c
ErikReider: Good point! We should check though why we have two fields before removing it though :)

types/scene/wlr_scene.c
ErikReider: Good catch!

types/scene/wlr_scene.c
ErikReider: The node should always be "visible" for us to render the blur onto the texture.

In the "workspace_get_blur" function we check if there even are any nodes that want optimized blur. If not, we won't render it.

This is very janky though with it being a node and such so an alternative to this would be nice xD

We could maybe create add a field in the root called "optimized_blur_tree" (SwayFX would have the bottom layer tree) or something and just check each render itr if that whole tree has rendered and render the optimized blur?

Not sure how we would check which tree is currently rendering though

types/scene/wlr_scene.c
```diff
@@ -1868,6 +2078,51 @@ static bool scene_entry_try_direct_scanout(struct render_list_entry *entry,
 	return true;
 }
 
+static struct blur_info workspace_get_blur_info(int list_len,
```
ErikReider, line 2171: Maybe we can merge this with ‘construct_render_list_iter‘?

include/scenefx/types/wlr_scene.h
ErikReider: > Q related to above, why do we need global blur params?

https://github.com/wlrfx/scenefx/pull/33#discussion_r1829048340

include/scenefx/types/wlr_scene.h
WillPower3309: Oh you're correct - was referencing the wrong struct

types/scene/wlr_scene.c
WillPower3309: Will do as a follow-up since it also impacts regulars texture rendering

include/scenefx/types/wlr_scene.h
WillPower3309: Hmm - but couldn't we traverse the scene tree to find the surfaces below? whether or not it's a layer surface wouldn't be relevant. Unless the impl would have separate trees for each layer that aren't linked at all - not sure if we have one root tree that has sub trees per layer or if they exist independently

WillPower3309:
> Optimized blur (x-ray) example in tinywl is now added. It's **very** simple but it works good enough to demonstrate what it does and how it works :)
> 
> image
> 
> _Notice how only the white rectangle gets blurred 😉_

both rectangles now seem part of the blurring, other windows are not though so just seems like a layer order thing!

include/scenefx/types/wlr_scene.h
ErikReider: Wouldn't iterating through the tree an extra time bring some possible performance issues?

ErikReider:
> both rectangles now seem part of the blurring, other windows are not though so just seems like a layer order thing!

Fixed by [698373b](https://github.com/wlrfx/scenefx/pull/33/commits/698373bc19c0e494d8dc17d9fb19125427428590) :D

WillPower3309 (approved):

PR #35, opened 2024-03-23
feat: add functions required by SwayFX

WillPower3309:


WillPower3309:
Seems windows with CSD decorations are broken:


Also, shadow scaling needs to be fixed

ErikReider:
> Seems windows with CSD decorations are broken:
> 
> 
> Also, shadow scaling needs to be fixed

Should be a swayfx issue if I'm not mistaken. We're not providing the correct geometry to the renderer (xdg geometry for the xdg toplevels)

render/fx_renderer/fx_pass.c
```diff
@@ -480,7 +480,7 @@ void fx_render_pass_add_box_shadow(struct fx_gles_render_pass *pass,
 	struct wlr_box shadow_box = options->shadow_box;
 	assert(shadow_box.width > 0 && shadow_box.height > 0);
 	struct wlr_box surface_box = options->clip_box;
-	float blur_sigma = shadow_data->blur_sigma * options->scale;
+	float blur_sigma = shadow_data->blur_sigma;
```
WillPower3309: seems weird to scale this here when the comp would have already needed to scale it for damage tracking purposes. Open to thoughts though, I know we scale elsewhere in the rendering functions here, but maybe we shouldn't.

ErikReider (approved):
LGTM! :)

PR #37, opened 2024-04-15
feat: complete nix flake

ozwaldorf:
## Summary

Refactor the nix flake to a complete version, following the same principles of the recent swayfx flake overhaul.

#### Packages

Output a default and scenefx package, which provides `out` and `lib` outputs (compatible with pkg-config)

#### Overlays

Provides both an insert and override overlay, one that will work with any nixpkgs and just insert the locked package as is, or the more advanced override that must be used with a smiliar nixpkgs unstable version, which can be used to overlay additional things etc.

---

> [Passing Garnix CI](https://github.com/ozwaldorf/scenefx/runs/23842110722)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -66,21 +66,15 @@
       devShells = forEachSystem (pkgs: {
         default = pkgs.mkShell {
           name = "scenefx-shell";
-
           inputsFrom = [
             self.packages.${pkgs.system}.scenefx
             pkgs.wlroots_0_17
           ];
-
-          nativeBuildInputs = [
-            pkgs.hwdata # for wlroots
-          ];
-
-          shellHook = with pkgs; ''
+          shellHook = ''
             (
-              mkdir -p "$PWD/subprojects"
-              cd "$PWD/subprojects"
-              cp -R --no-preserve=mode,ownership ${wlroots_0_17.src} wlroots
+              # Copy the nix version of wlroots into the project
+              mkdir -p "$PWD/subprojects" && cd "$PWD/subprojects"
+              cp -R --no-preserve=mode,ownership ${pkgs.wlroots_0_17.src} wlroots
             )'';
         };
       });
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -1,49 +1,90 @@
...
+          name = "scenefx-shell";
+
+          inputsFrom = [
+            self.packages.${pkgs.system}.scenefx
+            pkgs.wlroots_0_17
+          ];
+
+          nativeBuildInputs = [
+            pkgs.hwdata # for wlroots
+          ];
```
RicArch97: It looks like wlroots 0.17 in nixpkgs already has that input, do we still need it here, if we already have `wlroots_0_17` in `inputsFrom`?

https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/libraries/wlroots/default.nix#L132

RicArch97 (commented):
Looks good, just one comment

flake.nix
ozwaldorf: Maybe not, i carried it over from the legacy code since it had the explicit comment. If it works without I'm happy to make the change!

RicArch97: Ah ye, i see Will added that in the SwayFX flake including the comment, so @WillPower3309 what do you think?

WillPower3309: Yeah for whatever reason I need to explicitly add it or swayfx's wlroots subproject won't compile. I'll try it with and without on this branch and see if it's required here too

ozwaldorf:
Also, seems like a better fix for the headers is included separately in #39, so once merged will rebase this pr on top

flake.nix
WillPower3309: can confirm this block isn't needed. Weird!

WillPower3309: otherwise LGTM

ozwaldorf: Cleaned up the devshell!

RicArch97 (approved):
LGTM

WillPower3309 (approved):

PR #42, opened 2024-04-18
Fix: blur effects

### LINKED ISSUE #41: Blur causes purple tint to originally black background
Found when testing in combination with SwayFX.

Setting `blur_saturation 1` seems to resolve that

Blur settings active:
```
blur enable
blur_passes 3
```

ozwaldorf:
## Summary

- The combined transform matrix is currently transposed when multiplying against the color, resulting in some odd color shifts when using saturation. Fixed and left an explicit comment for the next person
- The current rng/noise function used has some issues on nvidia resulting in some line artifacts with higher noise values. Replaced with a new function used by hyprland in https://github.com/hyprwm/Hyprland/pull/5607. Note that the conditional for float precision is removed and always set to high due to nvidia not working correctly on medium precision.
- Be consistent in the ordering of [brightness, contrast, saturation, noise]

## Related issues

Fixes #41

### DIFF
--- a/render/fx_renderer/gles2/shaders/blur_effects.frag
+++ b/render/fx_renderer/gles2/shaders/blur_effects.frag
@@ -51,7 +51,9 @@ float hash(vec2 p) {
 
 void main() {
 	vec4 color = texture2D(tex, v_texcoord);
-	color *= brightnessMatrix() * contrastMatrix() * saturationMatrix();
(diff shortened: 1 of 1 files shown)


RicArch97:
Also: fixes #41

WillPower3309 (approved):
LGTM! Thanks for this

WillPower3309 (approved):

PR #56, opened 2024-08-14
Remove our buffer clipping logic

ErikReider:
The user should use wlr_scene_subsurface_tree_set_clip instead.

Causes some weird glitching when resizing GTK4 applications


WillPower3309 (approved):
LGTM!

PR #57, opened 2024-09-03
fix: add wayland-scanner to native build dependencies

eerii:
Build is broken on unstable nixpkgs after NixOS/nixpkgs#214906 has been merged. `wayland-scanner` is a different package now and it needs to be explicitly added to the native build dependencies.


WillPower3309:
Thanks!

PR #61, opened 2024-09-18
feat: removed stencil from shadow rendering

WillPower3309:


WillPower3309:
Shadow portion of #62

WillPower3309:
Hmm, strangely enough it seems that there is a small gap between the windows corner (when rounded) and the shadows corner, which is strange since we'd use the same roundRectSDF function :thinking:

render/fx_renderer/fx_pass.c
```diff
@@ -516,19 +470,16 @@ void fx_render_pass_add_box_shadow(struct fx_gles_render_pass *pass,
...
+	glUniform1f(renderer->shaders.box_shadow.blur_sigma, shadow_data->blur_sigma);
 	glUniform1f(renderer->shaders.box_shadow.corner_radius, options->corner_radius);
 	glUniform2f(renderer->shaders.box_shadow.size, shadow_box.width, shadow_box.height);
 	glUniform2f(renderer->shaders.box_shadow.position, shadow_box.x, shadow_box.y);
 
-	render(&shadow_box, &render_region, renderer->shaders.box_shadow.pos_attrib);
-
-	pixman_region32_fini(&render_region);
+	// TODO: properly remove area we don't need to be damaged
+	render(&shadow_box, options->clip, renderer->shaders.box_shadow.pos_attrib);
```
ErikReider: We need this before merging. I'll have a crack at this while your timezone is sleeping if you don't mind :D

ErikReider (changes requested):
Otherwise, LGTM! The improvement really is noticable :D

ErikReider (approved):

PR #63, opened 2024-09-18
Fixed shadows not respecting offset

ErikReider:
Made an oopsie in #61. Forgot to check if the shadow offset worked as expected (which wasn't the case 😅)


WillPower3309:
Merging for now but I have an idea to improve this down the line

WillPower3309 (approved):

PR #66, opened 2024-10-18
feat: add shadow_node

WillPower3309:


render/fx_renderer/fx_pass.c
```diff
@@ -453,48 +452,29 @@ void fx_render_pass_add_rounded_border_corner(struct fx_gles_render_pass *pass,
...
+	struct wlr_box box = options->box;
+	assert(box.width > 0 && box.height > 0);
 
 	push_fx_debug(renderer);
-
 	// blending will practically always be needed (unless we have a madman
 	// who uses opaque shadows with zero sigma), so just enable it
 	setup_blending(WLR_RENDER_BLEND_MODE_PREMULTIPLIED);
-	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA); // TODO: needed?
```
ErikReider: Needed, otherwise the alpha ins't applied properly. We should also add this to the rect functions bc their opacity doesn't work properly rn

render/fx_renderer/fx_pass.c
```diff
@@ -453,48 +452,29 @@ void fx_render_pass_add_rounded_border_corner(struct fx_gles_render_pass *pass,
...
-	pixman_region32_init(&render_region);
-
-	pixman_region32_t inner_region;
-	pixman_region32_init_rect(&inner_region,
-			surface_box.x + options->corner_radius * 0.3,
-			surface_box.y + options->corner_radius * 0.3,
-			fmax(surface_box.width - options->corner_radius * 0.6, 0),
-			fmax(surface_box.height - options->corner_radius * 0.6, 0));
-	pixman_region32_subtract(&render_region, options->clip, &inner_region);
-	pixman_region32_fini(&inner_region);
```
ErikReider, line 474: How do we exclude the inner section?

types/scene/wlr_scene.c
```diff
@@ -246,6 +251,8 @@ static void scene_node_opaque_region(struct wlr_scene_node *node, int x, int y,
 		if (scene_rect->color[3] != 1) {
 			return;
 		}
+	} else if (node->type == WLR_SCENE_NODE_SHADOW) {
+		// TODO
```
ErikReider: Should always return due to the shadow never being opaque

types/scene/wlr_scene.c
```diff
@@ -258,10 +265,7 @@ static void scene_node_opaque_region(struct wlr_scene_node *node, int x, int y,
 		}
 
 		if (scene_buffer->corner_radius > 0) {
-			return;
-		}
-
-		if (scene_buffer_has_shadow(&scene_buffer->shadow_data)) {
+			// TODO: this is incorrect
```
ErikReider, line 269: How so?

types/scene/wlr_scene.c
```diff
@@ -1231,6 +1258,23 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+			.box = dst_box,
+			.blur_sigma = scene_shadow->blur_sigma,
+			.corner_radius = scene_shadow->corner_radius,
+			.color = {
+				.r = scene_shadow->color[0],
+				.g = scene_shadow->color[1],
+				.b = scene_shadow->color[2],
+				.a = scene_shadow->color[3],
+			},
+			.clip = &render_region,
```
ErikReider, line 1371: Shouldn't we also add a way of creating a hole for the toplevel, something like ignore_box and ignore_corner_radius?

ErikReider (changes requested):

render/fx_renderer/fx_pass.c
WillPower3309: I figured it was needed, but saw we didn't use it for rects and wondered if it was handled in another layer. Thanks!

types/scene/wlr_scene.c
WillPower3309: Technically if the shadow has a sigma of 0 and an opacity of 1 it would be opaque, was debating if we wanted to handle that improbable case

render/fx_renderer/fx_pass.c
WillPower3309: I need to dive deeper into how we should handle this given that any node could be over it since it's no longer tied to a buffer node

types/scene/wlr_scene.c
WillPower3309: Yup! Related to the ignoring of the damage, need to think about how I should expose this

types/scene/wlr_scene.c
WillPower3309: If the buffer has a corner radius but is otherwise opaque we should just return the corner regions

WillPower3309:
remaining work:
+ [ ] render around a defined wlr_box to support offset (or maybe we just add the offset back - that may be better actually)
+ [ ] properly size and position the border node -> we shouldn't need the comp to take blur_sigma into account when defining the shadow width and height. See the scene-graph example for size / position of the shadow node to see what I mean
+ [ ] update tinywl

WillPower3309:
My tinywl work is barely functional, will need more work

tinywl/tinywl.c
```diff
@@ -709,7 +705,9 @@ static void xdg_toplevel_map(struct wl_listener *listener, void *data) {
 
 	wl_list_insert(&toplevel->server->toplevels, &toplevel->link);
 
-	focus_toplevel(toplevel, toplevel->xdg_toplevel->base->surface);
+	struct wlr_surface *surface = toplevel->xdg_toplevel->base->surface;
+	wlr_scene_shadow_set_size(toplevel->shadow, surface->current.width, surface->current.height); // TODO: use get node size?
```
ErikReider: This should be the xdg geometry. The surface size would include potential csd dimensions

tinywl/tinywl.c
```diff
@@ -855,9 +859,11 @@ static void server_new_xdg_surface(struct wl_listener *listener, void *data) {
 	/* Set the scene_nodes decoration data */
 	toplevel->opacity = 1;
 	toplevel->corner_radius = 20;
-	toplevel->shadow_data = shadow_data_get_default();
-	toplevel->shadow_data.enabled = true;
-	toplevel->shadow_data.color = (struct wlr_render_color) {1.0f, 0.0f, 0.0f, 1.0f};
+
+	float blur_sigma = 20.0f;
+	toplevel->shadow = wlr_scene_shadow_create(&toplevel->server->scene->tree,
```
ErikReider: Set this to toplevel->scene_tree so that the xdg surface and the shadow node share the same parent. By doing this you should be able to set the shadow position once (x/y: -blur_radius) ;)

tinywl/tinywl.c
```diff
@@ -786,6 +784,12 @@ static void xdg_toplevel_request_move(
 	 * client, to prevent the client from requesting this whenever they want. */
 	struct tinywl_toplevel *toplevel = wl_container_of(listener, toplevel, request_move);
 	begin_interactive(toplevel, TINYWL_CURSOR_MOVE, 0);
+
+	// TODO
+	struct wlr_box surface_box;
+	wlr_xdg_surface_get_geometry(toplevel->xdg_toplevel->base, &surface_box);
+	wlr_scene_node_set_position(&toplevel->shadow->node, surface_box.x - toplevel->shadow->blur_sigma,
```
ErikReider: With the suggested changes in the other comment, this shouldn't be needed due to the positions being parent relative.

I.e. parent tree pos = (30, 30), the shadow tree pos of (0, 0) would be (30, 30)

types/scene/wlr_scene.c
```diff
@@ -1231,6 +1278,41 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 		};
 		fx_render_pass_add_rect(data->render_pass, &rect_options);
 		break;
+	case WLR_SCENE_NODE_SHADOW:;
+		struct wlr_scene_shadow *scene_shadow = wlr_scene_shadow_from_node(node);
+
+		struct wlr_box window_box;
+		int window_x = 0, window_y = 0, window_width = 0, window_height = 0, window_corner_radius = 0;
+		struct wlr_scene_node *parent_node = wl_container_of(node->link.prev, parent_node, link);
+		if (parent_node->type == WLR_SCENE_NODE_BUFFER || parent_node->type == WLR_SCENE_NODE_RECT) {
```
ErikReider: This should preferably be done in a loop to get all of the above siblings :)

types/scene/wlr_scene.c
```diff
@@ -1231,6 +1278,41 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+			scene_node_get_size(node, &window_width, &window_height);
+			scene_node_get_corner_radius(node, &window_corner_radius);
+			printf("found window box %d, %d: width: %d, height %d\n", window_x, window_y, window_width, window_height);
+		}
+		window_box = (struct wlr_box) {
+		    .x = window_x,
+			.y = window_y,
+			.width = window_width,
+			.height = window_height
+		};
```
ErikReider: We should probably also transform the box

include/scenefx/types/wlr_scene.h
```diff
@@ -140,6 +140,15 @@ struct wlr_scene_rect {
 	float color[4];
 };
 
+/** A scene-graph node displaying a shadow */
+struct wlr_scene_shadow {
+	struct wlr_scene_node node;
+	int width, height;
+	int corner_radius;
+	float color[4];
```
ErikReider, line 148: Might make our lives a little easier if we use the wlr_color struct

ErikReider (changes requested):

tinywl/tinywl.c
WillPower3309: Yup, left a comment about that. I believe the get_node_size function does all that for us, plus works with rects as well

types/scene/wlr_scene.c
WillPower3309: Would we want to do this? If a sibling is transparent we should see the main surface's shadow beneath it

types/scene/wlr_scene.c
WillPower3309: Yup, need to add that as a Todo before this one is ready

include/scenefx/types/wlr_scene.h
WillPower3309: Similar to the rect struct we use it behind the scenes

WillPower3309: Pun intended

types/scene/wlr_scene.c
ErikReider: > Would we want to do this? If a sibling is transparent we should see the main surface's shadow beneath it

I meant to check all siblings, not just the first sibling

WillPower3309: I may have a gap here, what's the behaviour that we'd like to avoid that would result from not checking those? The first sibling would be the buffer or rect the shadow is drawn around

ErikReider: > I may have a gap here, what's the behaviour that we'd like to avoid that would result from not checking those? The first sibling would be the buffer or rect the shadow is drawn around

It depends on the layout, but we'd check all enabled nodes and add all of those to a larger rect that can include borders and surfaces (assuming that the borders and surface is above the shadow)

WillPower3309:
Updated remaining work comment

WillPower3309:
We can support multiple "window boxes" as follow-up work. We can use https://www.lighthouse3d.com/tutorials/glsl-tutorial/uniform-variables/ as a handy reference here

types/scene/wlr_scene.c
WillPower3309: Will do as follow-up, for now will rely on the next node

ErikReider (approved):
Everything seems to work as expected, LGTM! :)

PR #68, opened 2024-10-25
Minor fixes

ErikReider:
Fixes:
- Shadow inner region sometimes being able to become negative if the inner size is small enough
- Fixes rect alpha not working


render/fx_renderer/fx_pass.c
```diff
@@ -356,13 +356,17 @@ void fx_render_pass_add_rect(struct fx_gles_render_pass *pass,
 	push_fx_debug(renderer);
 	setup_blending(color->a == 1.0 ? WLR_RENDER_BLEND_MODE_NONE : options->blend_mode);
 
+	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
```
WillPower3309, line 359: Upstream doesn't set this here. Going to peek into their impl

WillPower3309:
covered in #69

Issue #70, opened 2024-10-26
Move internal corner rendering to separate shader


WillPower3309:
We can get some minor perf improvements by clipping the entire window box from our decoration node rendering, and then render the four corners of the windows with the decoration blend shaders

WillPower3309:
We could move external corner rounding to a separate shader as well! This is coupled with #72

ErikReider:
Can we close this as well?

WillPower3309:
Not yet, I want to play with this first :) I have some optimization ideas

WillPower3309:
> Not yet, I want to play with this first :) I have some optimization ideas
Sort of a two step approach, where we render the internal corners in a separate step so we can take that perf hit of accounting for those corners in a smaller region

PR #71, opened 2024-10-26
feat: add CORNER_LOCATION enum + round specific corner functionality

ErikReider:
Compositors can now specify which corners of a scene_buffer should be rounded instead of all.

Usage:

```c
// Only round the bottom corners
wlr_scene_buffer_set_corner_radius(buffer, 12, CORNER_LOCATION_BOTTOM_LEFT | CORNER_LOCATION_BOTTOM_RIGHT);
```

image

This also fixes specifying corners on transformed outputs being broken.

image


WillPower3309:
This would mess up decoration rendering (unless it's paired with a round titlebar), but I have an idea of how to fix it. Will be done so as part of #70

render/fx_renderer/gles2/shaders/tex.frag
```diff
@@ -55,8 +59,26 @@ void main() {
...
+		float left = step(relative_pos.x - radius, 0.0);
+		float right = step(size.x - radius, relative_pos.x);
+		float bottom = step(size.y - radius, relative_pos.y);
+
+		float top_left = top * left * float(round_top_left);
+		float top_right = top * right * float(round_top_right);
+		float bottom_left = bottom * left * float(round_bottom_left);
+		float bottom_right = bottom * right * float(round_bottom_right);
+		// A value of 1 means that we're allowed to round at this position
+		float round_corners = top_left + top_right + bottom_left + bottom_right;
```
WillPower3309: lets move this to a function in its own file (similar to how we do for the gradient shader and how we will handle the round rect SDF) - then we can share it between this shader and the round rect shader

render/fx_renderer/fx_pass.c
```diff
@@ -374,19 +384,19 @@ void fx_render_pass_add_rounded_rect(struct fx_gles_render_pass *pass,
...
-	case TOP_RIGHT:
+	case CORNER_LOCATION_TOP_RIGHT:
 		shader = &renderer->shaders.quad_round_tr;
 		break;
-	case BOTTOM_LEFT:
+	case CORNER_LOCATION_BOTTOM_LEFT:
 		shader = &renderer->shaders.quad_round_bl;
 		break;
-	case BOTTOM_RIGHT:
+	case CORNER_LOCATION_BOTTOM_RIGHT:
```
WillPower3309, line 440: I think this breaks upstream -git. We can add this as a follow-up, since we'd also probably want to update the shader anyway. We should handle the individual corner rounding for the rects the same as we do the tex

WillPower3309:
clever solution! I think we may want to break this down a bit better though in order to ensure swayfx-git compat. We can add the missing pieces as part of the swayfx rebase

render/fx_renderer/fx_pass.c
WillPower3309: actually let's just update upstream!

WillPower3309 (approved):

Issue #73, opened 2024-11-08
Shadow mask not scaling with output


ErikReider:
Tested with output scale set to 1.5:

![image](https://github.com/user-attachments/assets/b1836e82-2da5-4376-8bdc-1e80b90229f5)

PR #75, opened 2024-11-10
Fixed shadow mask not being scaled

ErikReider:
Fixes: #73


types/scene/wlr_scene.c
```diff
@@ -1389,6 +1389,8 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 		int window_corner_radius;
 		scene_get_next_sibling_geometry(node, &window_box, &window_corner_radius);
 
+		window_corner_radius *= data->scale;
+		scale_box(&window_box, data->scale);
```
WillPower3309, line 1395: Should we also do this for the round rect rendering above?

WillPower3309 (approved):

PR #79, opened 2024-11-17
Fixed sibling window rects not compensating for output position offsets

ErikReider:
When the ouptut position offset is set to something larger than 0, the sibling window coords automatically uses that offset, which isn't necessary.

image

Tested the fix in tinywl and fx-comp

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -1491,7 +1491,7 @@ static void get_tree_geometry(struct wlr_scene_node *node,
 }
 
 static void scene_get_next_sibling_geometry(struct wlr_scene_node *node,
-		struct wlr_box *box, int *corner_radius) {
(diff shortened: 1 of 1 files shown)


types/scene/wlr_scene.c
```diff
@@ -1589,6 +1589,9 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 			int window_corner_radius;
 			scene_get_next_sibling_geometry(node, &window_box, &window_corner_radius);
 
+			window_box.x -= data->logical.x;
+			window_box.y -= data->logical.y;
```
WillPower3309: Lgtm, but should we move this to the get sibling geometry function?

ErikReider: Done

WillPower3309 (approved):

PR #82, opened 2024-11-24
removed unused border corner shaders

WillPower3309:
Now that round rects should be used as our rounded window borders, we no longer need these shaders


ErikReider (approved):
LGTM!

PR #84, opened 2024-12-07
swayfx compat changes

WillPower3309:
This PR contains changes to scenefx related to the swayfx rebase (https://github.com/WillPower3309/swayfx/pull/367)

### DIFF
--- a/render/fx_renderer/fx_framebuffer.c
+++ b/render/fx_renderer/fx_framebuffer.c
@@ -69,9 +69,12 @@ GLuint fx_framebuffer_get_fbo(struct fx_framebuffer *buffer) {
 }
 
 void fx_framebuffer_get_or_create_custom(struct fx_renderer *renderer,
-		struct wlr_output *output, struct fx_framebuffer **fx_framebuffer) {
+		struct wlr_output *output, struct wlr_swapchain *swapchain,
+		struct fx_framebuffer **fx_framebuffer) {
 	struct wlr_allocator *allocator = output->allocator;
-	struct wlr_swapchain *swapchain = output->swapchain;
+	if (!swapchain) {
+		swapchain = output->swapchain;
+	}
 	int width = output->width;
 	int height = output->height;
 	struct wlr_buffer *wlr_buffer = NULL;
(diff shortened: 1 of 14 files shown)


types/scene/wlr_scene.c
```diff
@@ -2421,6 +2421,11 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
 		timer->pre_render_duration = timespec_to_nsec(&duration);
 	}
 
+	// TODO: we need to revisit this, this is a stopgap to ensure we can resume swayfx rebase
```
WillPower3309: this feels wrong. We should sanity check this with scene upstream. How do they handle the case of output swapchain not being enabled here? Are we missing an initialization on the swayfx side?

render/fx_renderer/fx_framebuffer.c
```diff
@@ -69,14 +69,20 @@ GLuint fx_framebuffer_get_fbo(struct fx_framebuffer *buffer) {
...
+	if (!swapchain) {
+		swapchain = output->swapchain;
+	}
 	int width = output->width;
 	int height = output->height;
 	struct wlr_buffer *wlr_buffer = NULL;
 
 	if (*fx_framebuffer == NULL) {
+		if (swapchain == NULL) {
+			printf("swapchain is null\n");
```
ErikReider: Return early here?

render/fx_renderer/fx_renderer.c
```diff
@@ -565,6 +568,7 @@ struct wlr_renderer *fx_renderer_create_egl(struct wlr_egl *egl) {
 	wlr_egl_unset_current(renderer->egl);
 
 	get_fx_shm_formats(renderer, &renderer->shm_texture_formats);
+	printf("made renderer\n");
```
ErikReider: We should use wlr_log instead

WillPower3309: Oop that was a debugging print I think

render/fx_renderer/fx_framebuffer.c
WillPower3309: Oops must've pushed a debug change

types/scene/wlr_scene.c
ErikReider: It was due to testing the new output modeset before applying if I'm not mistaken. So a new non-output swapchain is created and passed to scene. There's already a swapchain check in the output_build_state function so that should be good enough :)

WillPower3309:
This has some good commits for mainline. Going to merge and can open new PRs for any other needed work

WillPower3309:
Ill let @ErikReider give an approval first though in case he thinks this needs anything else

ErikReider (approved):
LGTM! I don't really have any reference about the gradient logic but those changes don't look out of place :)

PR #86, opened 2025-01-09
Replaced sibling clip region detection

ErikReider:
Replaces the automatic hole detection logic with a more controlled struct which needs corner radius and size of a box to clip.

TODO:
- [x] wlr_scene.h function documentation
- [x] Use a different word than "hole"?
- [x] Works with multiple monitors
- [x] Scaling tested
- [x] Output rotation tested

Swayfx changes to get this working:
```diff
diff --git a/sway/desktop/transaction.c b/sway/desktop/transaction.c
index 215131d7..ece6f5c6 100644
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -448,6 +448,16 @@ static void arrange_container(struct sway_container *con,
 					width + config->shadow_blur_sigma * 2,
 					height + config->shadow_blur_sigma * 2);
 
+			wlr_scene_shadow_set_hole_data(con->shadow, (struct hole_data) {
+					.corner_radius = con->corner_radius,
+					.size = {
+						.x = 0,
+						.y = 0,
+						.width = width,
+						.height = height,
+					}
+			});
+
 			int x = config->shadow_offset_x - config->shadow_blur_sigma;
 			int y = config->shadow_offset_y - config->shadow_blur_sigma;
 			wlr_scene_node_set_position(&con->shadow->node, x, y);
```


WillPower3309:
> Use a different word than "hole"?

100% xD I can't think of one at the moment though

WillPower3309:
could use terminology like `clipped_region`?

WillPower3309:
some bugs here I think. When the hole is set to the correct corner radius:

when the corner radius is increased:

ErikReider:
> some bugs here I think. When the hole is set to the correct corner radius:
> 
> when the corner radius is increased:

Works as expected without borders, so we might need to compensate for the sway borders in the swayfx patch

WillPower3309:
> some bugs here I think. When the hole is set to the correct corner radius:  when the corner radius is increased: 

also fixed in fixed in https://github.com/wlrfx/scenefx/pull/88

render/fx_renderer/gles2/shaders/box_shadow.frag
```diff
@@ -80,7 +87,10 @@ void main() {
     shadow_alpha += (random() - 0.5) / 128.0;
 
     // get the window alpha so we can render around the window (fix pixel gap by adding 1.0 to radius)
-    float window_alpha = smoothstep(-1.0, 1.0, roundRectSDF(window_half_size, window_position, window_corner_radius + 1.0));
+    float window_corner_alpha = hole_alpha(window_size, window_position,
+            window_corner_radius,
+            window_round_top_left, window_round_top_right,
+            window_round_bottom_left, window_round_bottom_right);
```
WillPower3309: why not reuse the corner_alpha function? why do we need a new hole_alpha shader?

WillPower3309 (commented):
Finally got around to looking closer at this, one comment

render/fx_renderer/gles2/shaders/box_shadow.frag
WillPower3309: I realize I fixed up that function after this PR was created, but I'd think we can just reuse it here like we had roundRectSDF

WillPower3309:
We can also probably get rid of roundRectSDF with this change, since we don't use it anymore

render/fx_renderer/gles2/shaders/box_shadow.frag
ErikReider: The reason why I created the new function is due to the failsafes that the hole needs. Without it, corners set to none while a corner radius is set would create a wonky output.

But I could copy/paste that logic into the rounded quad shader as well

WillPower3309: I think we should safeguard both shaders in that case, they should be the same

WillPower3309:
Is documentation all that is left? I can try to get this merged today after work

ErikReider:
> Is documentation all that is left? I can try to get this merged today after work

Ooops, forgot to check the box. Marking as ready for review! :D

render/fx_renderer/gles2/shaders/box_shadow.frag
```diff
@@ -68,6 +72,9 @@ float random() {
     return fract(sin(dot(vec2(12.9898, 78.233), gl_FragCoord.xy)) * 43758.5453);
 }
 
+float corner_alpha(vec2 size, vec2 position, float radius,
+            bool round_tl, bool round_tr, bool round_bl, bool round_br);
+
 float roundRectSDF(vec2 half_size, vec2 position, float radius);
```
ErikReider: Remove function

render/fx_renderer/shaders.c
```diff
@@ -242,9 +246,9 @@ bool link_tex_program(struct tex_shader *shader, enum fx_tex_shader_source sourc
 }
 
 bool link_box_shadow_program(struct box_shadow_shader *shader) {
-	GLchar shadow_src[4096];
-	snprintf(shadow_src, sizeof(shadow_src), "%s\n%s", box_shadow_frag_src,
-			round_rect_sdf_frag_src);
+	GLchar shadow_src[8192];
```
ErikReider, line 248: Lower to 4096 if possible

ErikReider: Still not possible, even after removing `roundRectSDF` :(

include/scenefx/types/fx/clipped_region.h
```diff
@@ -0,0 +1,16 @@
+#ifndef TYPES_FX_CLIPPED_REGION_H
+#define TYPES_FX_CLIPPED_REGION_H
+
+#include <wlr/util/box.h>
+
+#include "scenefx/types/fx/corner_location.h"
+
+struct clipped_region {
+	struct wlr_box size;
```
WillPower3309: I think `area` would be more appropriate since that struct contains positional data as well

WillPower3309:
Hmm, getting a segfault on tinywl. Was present before my naming update

WillPower3309 (approved):

PR #98, opened 2025-01-29
Clipped region fixes

ErikReider:
- [x] Made positioning node-relative
- [x] Fixed clipped region not working with non-rounded rects
- [x] Added tinywl example
- [ ] ~Fix clipped region being wonky when all corners aren't rounded~ *Will will post a PR later*


WillPower3309:
Looks good so far! We'll have to get gradients up to speed as well, I'll create a tracking issue after this is merged.

WillPower3309 (approved):

PR #100, opened 2025-02-08
feat: per corner rounding (and small corner shader optimization)

WillPower3309:


render/fx_renderer/gles2/shaders/corner_alpha.frag
```diff
@@ -1,21 +1,19 @@
...
-	vec2 bottom_left = abs(relative_pos - vec2(size.x, 0)) - size + radius;
-	vec2 bottom_right = abs(relative_pos) - size + radius;
+	vec2 top_left = abs(relative_pos - size) - size + radius_tl;
+	vec2 top_right = abs(relative_pos - vec2(0, size.y)) - size + radius_tr;
+	vec2 bottom_left = abs(relative_pos - vec2(size.x, 0)) - size + radius_bl;
+	vec2 bottom_right = abs(relative_pos) - size + radius_br;
+
+	float dist = max(
+		max(get_dist(top_left, radius_tl), get_dist(top_right, radius_tr)),
+		max(get_dist(bottom_left, radius_bl), get_dist(bottom_right, radius_br))
```
WillPower3309, line 15: this can probably still be optimized considering we do the exact same operation on each value, the only val that changes is the `q` input, so we should be able to choose the proper q and then call `get_dist` once.

WillPower3309:
Kept the individual rounding to the shader level for now, we can experiment with exposing this to compositors as a follow-up, don't want to hold back the swayfx release any longer

ErikReider (approved):
Every test case that I've thrown at it is displaying correctly, so LGTM!

I'll let you merge in case you want a last try at optimizing the shader :)

PR #102, opened 2025-02-10
feat: add gradients to scene_rect

WillPower3309:
Relatively rough for now... Will likely need to clean up rect rendering as a follow-up, so much shared logic between the rounded vs non rounded vs gradient vs solid color

### DIFF
--- a/include/render/fx_renderer/shaders.h
+++ b/include/render/fx_renderer/shaders.h
@@ -46,10 +46,10 @@ struct quad_grad_shader {
 	GLint degree;
 	GLint grad_box;
 	GLint pos_attrib;
-	GLint linear;
+	GLint is_linear;
 	GLint origin;
 	GLint count;
-	GLint blend;
+	GLint should_blend;
 };
 
 bool link_quad_grad_program(struct quad_grad_shader *shader, GLint client_version, int max_len);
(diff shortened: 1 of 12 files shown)


WillPower3309:
should probably add an assertion for gradient `count`

WillPower3309:
an idea... we could treat all rects as having a gradient but only with one color in it if it is solid color

ErikReider:
> an idea... we could treat all rects as having a gradient but only with one color in it if it is solid color

Yeah I was just thinking that!! :D

WillPower3309:
bug with blending and conic gradient:


```c
	/* Add a bottom rect to demonstrate optimized blur */
	float bottom_rect_color[16] = { 1, 1, 1, 1,   1, 0, 1, 1,   0, 0, 1, 1 };
	struct wlr_scene_rect *bottom_rect = wlr_scene_rect_create(server.layers.bottom_layer,
			200, 200, bottom_rect_color);
	wlr_scene_rect_set_gradient(
		bottom_rect,
		(struct gradient) {
			.degree = 0.0,
			.range = (struct wlr_box) {
				.x = 0,
				.y = 0,
				.width = 200,
				.height = 200,
			},
			.origin = { 0.5, 0.5 },
			.is_linear = false,
			.should_blend = true,
			.count = 3,
			.colors = bottom_rect_color,
		}
	);
```

kociap:
What work needs to be done to have this merged?

WillPower3309:
> What work needs to be done to have this merged?

A big rebase primarily! The bug I mentioned needs addressing too

PR #106, opened 2025-02-14
Major optimized blur simplification + minor rect fixes

ErikReider:
- [ ] Needs thorough testing:
  - [x] Rotated outputs
  - [x] Scaled outputs
  - [x] Multiple outputs
  - [x] Non-optimized blur damage artifact prevention

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -543,6 +543,17 @@ static bool scene_node_update_iterator(struct wlr_scene_node *node,
 		int lx, int ly, void *_data) {
 	struct scene_update_data *data = _data;
 
+	if (node->type == WLR_SCENE_NODE_OPTIMIZED_BLUR) {
+		struct wlr_scene_optimized_blur *scene_blur = wlr_scene_optimized_blur_from_node(node);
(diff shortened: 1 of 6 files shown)


types/scene/wlr_scene.c
```diff
@@ -276,6 +276,12 @@ static void scene_node_opaque_region(struct wlr_scene_node *node, int x, int y,
 		if (scene_rect->color[3] != 1) {
 			return;
 		}
+		if (!wlr_box_empty(&scene_rect->clipped_region.area)) {
+			struct wlr_box *clipped = &scene_rect->clipped_region.area;
+			pixman_region32_intersect_rect(opaque, opaque, clipped->x, clipped->y,
+					clipped->width, clipped->height);
```
WillPower3309: wouldn't we want to do the opposite? The clipped region is not opaque

WillPower3309: I tested with the opposite and the rendering still seemed to work fine, I'll wait for you to respond though in case I'm missing something!

ErikReider: Good catch!

WillPower3309 (approved):

PR #109, opened 2025-02-24
simplified individual corner passing logic

WillPower3309:


ErikReider:
Will this include the per corner radii or is this ready for review? :)

ErikReider (approved):
Clean! Nice to avoid all of those corner test cases in the shaders :D

PR #112, opened 2025-03-27
added some bacis stuff for animations

dqrk0jeste:
idea here is that this is the logic thats going to be used by compositors anyway, and having these helpers will remove all this boilerplate.
compositor maintainers are free to implement them in any way they want (with output frame events or with timers for example), and just call appropriate functions to start, update and destroy them.

todos
- [ ] add documentation in the header
- [ ] add the example implementation to `tinywl` 
  - [ ] timers example
  - [ ] output frame events example

i dont know how would i go about adding both examples to `tinywl` tho, so maybe i could just add the first option, as thats what `mwc` does and what i am familiar with.

### DIFF
--- a/include/scenefx/types/fx/animations.h
+++ b/include/scenefx/types/fx/animations.h
@@ -1,0 +1,41 @@
+#ifndef TYPES_FX_ANIMATIONS_H
+#define TYPES_FX_ANIMATIONS_H
+
+#include <stdint.h>
+#include <wayland-util.h>
+#include <wlr/types/wlr_xdg_shell.h>
+
+#ifndef FX_ANIMATIONS_BAKED_POINTS_COUNT
+#define FX_ANIMATIONS_BAKED_POINTS_COUNT 256
+#endif
+
+struct fx_animation_curve {
+	double params[4];
+
+	double baked_points_x[FX_ANIMATIONS_BAKED_POINTS_COUNT], baked_points_y[FX_ANIMATIONS_BAKED_POINTS_COUNT];
+};
+
+struct fx_animation_curve *
+fx_animation_curve_create(double params[static 4]);
+
+void
+fx_animation_curve_destroy(struct fx_animation_curve *curve);
+
+struct fx_translate_animation {
+	struct fx_animation_curve *curve;
+	struct wlr_box start, end;
+
+	struct wlr_box current;
+	bool finished;
+};
+
+struct fx_translate_animation *
+fx_translate_animation_create(struct wlr_box start, struct wlr_box end, struct fx_animation_curve *curve);
+
+void
+fx_translate_animation_destroy(struct fx_translate_animation *animation);
+
+struct wlr_box
+fx_translate_animation_update(struct fx_translate_animation *animation, double progress);
+
+#endif
--- a/types/fx/animations.c
+++ b/types/fx/animations.c
@@ -1,0 +1,84 @@
+#include "scenefx/types/fx/animations.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+struct fx_animation_curve *
+fx_animation_curve_create(double params[static 4]) {
+	struct fx_animation_curve *curve = calloc(1, sizeof(*curve));
+
+	memcpy(curve->params, params, 4 * sizeof(double));
+
+	for(size_t i = 0; i < FX_ANIMATIONS_BAKED_POINTS_COUNT; i++) {
+		double t = (double)i / (FX_ANIMATIONS_BAKED_POINTS_COUNT - 1);
+
+		curve->baked_points_x[i] =  3 * t * (1 - t) * (1 - t) * params[0]
+			+ 3 * t * t * (1 - t) * params[2]
+			+ t * t * t;
+
+		curve->baked_points_y[i] = 3 * t * (1 - t) * (1 - t) * params[1]
+			+ 3 * t * t * (1 - t) * params[3]
+			+ t * t * t;
+	}
+
+	return curve;
+}
+
+void
+fx_animation_curve_destroy(struct fx_animation_curve *curve) {
+	free(curve);
+}
+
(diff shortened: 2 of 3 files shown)


WillPower3309:
I like the idea of providing some help to compositors for animations. I think the timers example would be more than adequate for tinywl!

WillPower3309:
What are your thoughts on allowing compositors to add their own `update` and `complete` function pointers? I'd think `complete` would be required

dqrk0jeste:
i have greatly improved on the initial idea; one of the things i did was add timers by default with user callbacks. i have been doing it locally in mwc, but forgot to push the code yesterday, so hopefully i will finish it today so we have more ground for discussion.

dqrk0jeste:
> What are your thoughts on allowing compositors to add their own `update` and `complete` function pointers? I'd think `complete` would be required. We can then remove the `finished` boolean, and this would let us keep as much of the animation structs as private as possible, to ensure compositors won't mess around with the process. We could even manage the animation timer itself and enforce that as the preferred way of implementing animations, unless there are distinct advantages to the output frame events method (I'd assume compositors would want as little latency in that path as possible).

you edited the response in the meantime, but i have already mostly answered things here.

dqrk0jeste:
this should be done now, i have also added an example to tinywl :)

include/scenefx/types/fx/animations.h
```diff
@@ -0,0 +1,52 @@
...
+typedef void (*fx_transform_animation_callback_func_t)(struct wlr_box current, bool done, void *user_data);
+
+/* tranform animations will transform the `start` box to the `end` box,
+ * calling `callback` on regular intervals with the current box contained in `current` */
+struct fx_transform_animation;
+
+/* creates a new transform animation */
+struct fx_transform_animation *
+fx_transform_animation_create(struct wlr_box start, struct wlr_box end, uint32_t duration,
+        struct fx_animation_curve *curve, fx_transform_animation_callback_func_t callback, void *user_data);
```
ErikReider: Nit: follow the coding style of all the other files :)

(line length, when to add a new line, etc...)

tinywl/tinywl.c
```diff
@@ -598,6 +603,10 @@ static void xdg_toplevel_commit(struct wl_listener *listener, void *data) {
 		return;
 	}
 
+	/* we dont take the commits from a client if its animating currently.
+	 * this should be handled better, but for showing animations off this is fine */
+	if(toplevel->popin_animation) return;
```
ErikReider, line 610: Nit: Space between `if` and `(`

include/scenefx/types/fx/animations.h
```diff
@@ -0,0 +1,52 @@
...
+
+struct fx_animation_curve *
+fx_animation_curve_create(double params[static 4]);
+
+void
+fx_animation_curve_destroy(struct fx_animation_curve *curve);
+
+typedef void (*fx_transform_animation_callback_func_t)(struct wlr_box current, bool done, void *user_data);
+
+/* tranform animations will transform the `start` box to the `end` box,
```
ErikReider: Spelling: "Transform"*

include/scenefx/types/fx/animations.h
```diff
@@ -0,0 +1,52 @@
...
+ * calling `callback` on regular intervals with the current box contained in `current` */
+struct fx_transform_animation;
+
+/* creates a new transform animation */
+struct fx_transform_animation *
+fx_transform_animation_create(struct wlr_box start, struct wlr_box end, uint32_t duration,
+        struct fx_animation_curve *curve, fx_transform_animation_callback_func_t callback, void *user_data);
+
+/* a user may destroy the animation at any momment, stopping the timer callbacks and freeing the resources.
```
ErikReider: Spelling: "moment"*

tinywl/tinywl.c
```diff
@@ -827,6 +887,29 @@ static void xdg_toplevel_map(struct wl_listener *listener, void *data) {
 	wl_list_insert(&toplevel->server->toplevels, &toplevel->link);
 
 	focus_toplevel(toplevel, toplevel->xdg_toplevel->base->surface);
+
+	/* we setup the popin animation, we start it from the middle of the screen, with a size of 1, 1 */
```
ErikReider: Spelling: "pop-in"*

types/fx/animations.c
```diff
@@ -0,0 +1,200 @@
...
+void
+fx_animation_manager_init(struct wl_display *display, struct wlr_scene *scene) {
+    manager.display = display;
+    manager.scene = scene;
+}
+
+struct fx_animation_curve {
+	double params[4];
+
+	double baked_points_x[FX_ANIMATIONS_BAKED_POINTS_COUNT], baked_points_y[FX_ANIMATIONS_BAKED_POINTS_COUNT];
```
ErikReider: Nit: Line length

types/fx/animations.c
```diff
@@ -0,0 +1,200 @@
...
+} manager;
+
+void
+fx_animation_manager_init(struct wl_display *display, struct wlr_scene *scene) {
+    manager.display = display;
+    manager.scene = scene;
+}
+
+struct fx_animation_curve {
+	double params[4];
```
ErikReider, line 25: It would be nice to know what these 4 parameters are, and what they do as a comment?

https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/cubic-bezier#parameters

dqrk0jeste:
i guess a `.clang-format` would be nice to have, but i will format the code manually.

but on the more important note, we may want to also add 'general purpose' animations (`struct fx_animation`) that would just provide `current` factor in the callback for things like fade-in/out.

what are your thought on it?

dqrk0jeste:
formatting issues should be resolved. i have also added some resources for bezier curves, but in a header file instead, as it seemed better that way.

PR #115, opened 2025-05-06
Decoration clipping adjustments

walldmtd:
These are some small shader adjustments to make rounded corners look a little more natural.

- Outside clipping for the tex shader is pulled in by 0.25px to minimize overlap with borders, and for slightly smoother antialiasing.
- Outside clipping for rounded quads is pulled in by 0.5px to match the inside and keep the width consistent on the corners.
- Inside clipping for box shadows is pulled in by 0.75px to remove gaps between it and the borders.

There is also a fix in `fx_pass.c` to make box shadows use their corner radius and clip to the border properly.

Corners before and after:
corner-before corner-after

Shadows before and after (transparent 10px border so the shadow is visible behind it):

shadow-before shadow-after

### DIFF
--- a/render/fx_renderer/fx_pass.c
+++ b/render/fx_renderer/fx_pass.c
@@ -609,6 +609,7 @@ void fx_render_pass_add_box_shadow(struct fx_gles_render_pass *pass,
 	glUniform1f(renderer->shaders.box_shadow.blur_sigma, options->blur_sigma);
 	glUniform2f(renderer->shaders.box_shadow.size, box.width, box.height);
 	glUniform2f(renderer->shaders.box_shadow.position, box.x, box.y);
+	glUniform1f(renderer->shaders.box_shadow.corner_radius, options->corner_radius);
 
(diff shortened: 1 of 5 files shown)


render/fx_renderer/fx_pass.c
```diff
@@ -609,6 +609,7 @@ void fx_render_pass_add_box_shadow(struct fx_gles_render_pass *pass,
 	glUniform1f(renderer->shaders.box_shadow.blur_sigma, options->blur_sigma);
 	glUniform2f(renderer->shaders.box_shadow.size, box.width, box.height);
 	glUniform2f(renderer->shaders.box_shadow.position, box.x, box.y);
+	glUniform1f(renderer->shaders.box_shadow.corner_radius, options->corner_radius);
```
WillPower3309, line 612: wow how did we miss this? good catch

WillPower3309 (approved):

WillPower3309:
Thanks!

PR #117, opened 2025-05-17
Add support for gles3 shaders

walldmtd:
This is meant to be a fix for float precision issues in shaders caused by graphics drivers implementing the bare minimum of the gles2 spec. gles3 requires high precision support which fixes that problem.

This needs a LOT of testing, ideally on some older hardware (with OpenGL version <4.3, basically anything made before 2012).

The idea is that since the shaders are compiled at runtime, the codebase can still use gl2.h and doesn't need to be converted to gles3. All that needs to change is adding gles3 shader files, creating an EGL context with the right client version, and linking the right shaders.

There's a lot going on here, so I'll try to explain each part:

`egl.c`:
- Tries to create a context with client version 3. If the graphics drivers don't support it, the creation will fail and it will fall back to version 2.

`shaders.h/c`:
- Adds `GLint client_version` as an input parameter for the program linking functions to tell them which shader version to use.
- Links gles3 or gles2 shaders based on the client version passed in
- For shaders that have a `#define`, the value is added using an extra string format before that formatted string is combined with other shaders.
  - `#version 300 es` needs to be the first line in the shader, so the `#define` can't be prepended to the shader for gles3.

`fx_renderer.c`, `fx_pass.c`:
- Queries the EGL context for the client version, and passes it through when linking shaders.

Shaders:
- Added a `gles3` folder for the new shaders
- The gles3 shaders are identical to the gles2 shaders for consistency, aside from in/out variables, float precision settings, and the `#version` indicators
- `smooth` is renamed to `smooth_fac` in `gradient.frag` (`smooth` is reserved in gles3)
- Shaders that have a `#define` have it moved inside the shaders themselves to be filled in with a string format (needed for the `#version` identifier to be the first line)
- Minor formatting/whitespace changes for consistency


WillPower3309:
Thanks a ton for this! I'll try to find some time to review it this weekend

PR #125, opened 2025-06-24
pruned blur_data

WillPower3309:


WillPower3309:
This is WIP

include/scenefx/types/wlr_scene.h
```diff
@@ -118,7 +117,15 @@ struct wlr_scene {
...
 
-		struct blur_data blur_data;
+		struct {
+			int num_passes;
+			int radius;
+			float noise;
+			float brightness;
+			float contrast;
+			float saturation;
+		} blur_data;
```
ErikReider, line 127: Instead of pruning the whole struct entirely, why not a `struct blur_data *blur_data`, would solve all the include problems, just define the struct above, just like `struct wlr_output` on line 34.

Nothing wrong with this, just seems like a lot of copying and pasting fields if we were to add a new one

PR #127, opened 2025-06-29
A bunch of blur artifact fixes

ErikReider:
Just notices a few artifacts in SwayNotificationCenter and Kitty. So after investigating, these pesky bugs have been here since the 0.1 release...

The first bug was that we didn't damage the surrounding area of a committed/updated node properly, and the second bug was that the artifact prevention system wasn't active for a few partially opaque scene nodes, like SwayNotificationCenter.

Now also, running with the `WLR_SCENE_DEBUG_DAMAGE=highlight` environment variable should display the extended damage :)

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -21,6 +21,7 @@
 #include "scenefx/render/fx_renderer/fx_effect_framebuffers.h"
 #include "scenefx/render/fx_renderer/fx_renderer.h"
 #include "scenefx/render/pass.h"
+#include "scenefx/types/fx/blur_data.h"
 #include "scenefx/types/fx/clipped_region.h"
 #include "scenefx/types/fx/corner_location.h"
 #include "types/wlr_output.h"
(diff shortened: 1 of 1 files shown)


WillPower3309:
> @WillPower3309 I also noticed that we don't render any blur for scene_buffers that are "single pixel buffers", but we can tackle that in a future PR

oof. Mind opening an issue so we don't forget?

types/scene/wlr_scene.c
```diff
@@ -39,6 +40,13 @@
 #define DMABUF_FEEDBACK_DEBOUNCE_FRAMES  30
 #define HIGHLIGHT_DAMAGE_FADEOUT_TIME   250
 
+#define SCENE_BUFFER_SHOULD_BLUR(scene_buffer, scene) \
+	scene_buffer->backdrop_blur && is_scene_blur_enabled(&scene->blur_data) && \
+		(!scene_buffer->buffer_is_opaque || scene_buffer->opacity < 1.0f)
```
WillPower3309, line 45: Is it possible for the buffer to be opaque and the opacity to be less than 1? Haven't checked but not sure if is opaque is for the texture we are given

types/scene/wlr_scene.c
```diff
@@ -39,6 +40,13 @@
...
+#define SCENE_RECT_SHOULD_BLUR(scene_rect, scene) \
+	scene_rect->backdrop_blur && is_scene_blur_enabled(&scene->blur_data) && \
+		scene_rect->color[3] != 1
+
```
WillPower3309, line 49: Nit: feels weird to check if opacity is less than 1 above and not equal to 1 here, let's keep it consistent and check if it's less than 1

types/scene/wlr_scene.c
```diff
@@ -355,6 +363,7 @@ struct scene_update_data {
 #endif
 
 	bool optimized_blur_dirty;
+	struct wlr_scene *scene;
```
WillPower3309: We can just add the blur data instead of the whole scene struct

types/scene/wlr_scene.c
ErikReider: Yeah, the opacity is the one that compositors can set, like in SwayFX, while `is_opaque` indicates if the whole buffer is opaque or not.

Edit: Ok double-checked. It get's set when a surface commits a new buffer with the value from `wlr_buffer_is_opaque`:
Doc: `When true is returned, the buffer is guaranteed to be fully opaque, but the reverse is not true: false may be returned in cases where the buffer is fully opaque.`

WillPower3309 (approved):

PR #129, opened 2025-06-30
Add optional Tracy support

ErikReider:
Adds support for [tracy](https://github.com/wolfpld/tracy) which is used for advanced profiling! :D

Obligatory screenshot:
image

How to test it with tinywl, just add the `-Dtracy_enable=true` build option:
```sh
# scenefx/
meson setup build --buildtype=debugoptimized -Dtracy_enable=true --wipe
meson compile -C build
./build/tinywl/tinywl -s foot
```

And launch the separate tracy executable, which needs to be installed/compiled separately. The latest Wayland build is pretty broken on my system (crashes a lot...), so I compiled it with LEGACY=1, and run it through a separate terminal:

```sh
# tracy/
cmake -B profiler/build -S profiler -DCMAKE_BUILD_TYPE=Release -DLEGACY=1
cmake --build profiler/build --config Release --parallel
./profiler/build/tracy-profiler
```

Without the `tracy_enable` build option enabled (the default state), tracy support is completely disabled, so no overhead.

To test it with SwayFX, apply this [patch](https://gist.github.com/ErikReider/7a507d489974a83417a8d3984cb52248) and use the same build option to meson when setting up the project.

TODO:
- [ ] Test fewer/more functions?
- [ ] Add more CPU profiling to the wlr_scene functions
- [ ] Use custom names instead of function names?
- [ ] Optional that can be added here/at a later date:
	- [ ] Add `TracyCAppInfo`?
	- [ ] Add messages?
	- [ ] Add Graphs?

Note: There's a `tracy.wrap` file in the subprojects directory which only gets automatically cloned when tracy is enabled through Meson.

### DIFF
--- a/render/fx_renderer/fx_pass.c
+++ b/render/fx_renderer/fx_pass.c
@@ -14,6 +14,7 @@
 #include "render/fx_renderer/fx_renderer.h"
 #include "render/fx_renderer/shaders.h"
 #include "render/pass.h"
+#include "render/tracy.h"
 #include "scenefx/render/fx_renderer/fx_renderer.h"
 #include "scenefx/render/fx_renderer/fx_effect_framebuffers.h"
 #include "scenefx/types/fx/corner_location.h"
@@ -70,6 +71,7 @@ static bool render_pass_submit(struct wlr_render_pass *wlr_pass) {
 	struct fx_render_timer *timer = pass->timer;
 	bool ok = false;
 
+	TRACY_BOTH_ZONES_START(pass->buffer->renderer);
 	push_fx_debug(renderer);
 
 	if (timer) {
@@ -111,6 +113,10 @@ out:
 	glBindFramebuffer(GL_FRAMEBUFFER, 0);
 
 	pop_fx_debug(renderer);
+	TRACY_BOTH_ZONES_END;
+
+	TRACY_GPU_ZONE_COLLECT(renderer);
+
 	wlr_egl_restore_context(&pass->prev_ctx);
 
 	wlr_drm_syncobj_timeline_unref(pass->signal_timeline);
@@ -121,6 +127,7 @@ out:
 	}
 	free(pass);
 
+	TRACY_MARK_FRAME_END("Frame");
 	return ok;
 }
 
@@ -317,24 +324,28 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 	src_fbox.width /= options->texture->width;
 	src_fbox.height /= options->texture->height;
 
+	TRACY_BOTH_ZONES_START(renderer);
 	push_fx_debug(renderer);
 
 	if (options->wait_timeline != NULL) {
 		int sync_file_fd =
 			wlr_drm_syncobj_timeline_export_sync_file(options->wait_timeline, options->wait_point);
 		if (sync_file_fd < 0) {
+			TRACY_BOTH_ZONES_END;
 			return;
 		}
 
 		EGLSyncKHR sync = wlr_egl_create_sync(renderer->egl, sync_file_fd);
 		close(sync_file_fd);
 		if (sync == EGL_NO_SYNC_KHR) {
+			TRACY_BOTH_ZONES_END;
 			return;
 		}
 
 		bool ok = wlr_egl_wait_sync(renderer->egl, sync);
 		wlr_egl_destroy_sync(renderer->egl, sync);
 		if (!ok) {
+			TRACY_BOTH_ZONES_END;
 			return;
 		}
 	}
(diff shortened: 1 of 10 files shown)


render/fx_renderer/fx_pass.c
```diff
@@ -859,6 +982,40 @@ static struct fx_framebuffer *get_main_buffer_blur(struct fx_gles_render_pass *p
 	// Artifacts with NEAREST filter
 	fx_options->tex_options.base.filter_mode = WLR_SCALE_FILTER_BILINEAR;
 
+	TRACY_BOTH_ZONES_START(renderer);
+	TRACY_ZONE_TEXT_f("dst_box (WxH, X, Y): %ix%i, %i, %i",
```
WillPower3309: this goes for all tracy functions, but why use %i over %d? is this something specific to tracy? (I know scanf for example treats %i different than %d)

ErikReider: lol, I didn't know that! I'm on it :D

WillPower3309:
Hey @ErikReider is this ready? Only thing I see is that we need to add those `#if TRACY_ENABLE`s

ErikReider:
> Hey @ErikReider is this ready? Only thing I see is that we need to add those `#if TRACY_ENABLE`s

Oh, sorry, didn't see your original comment about that. There's one in the tracy.h file which is why there are so many macros. Makes it a lot more clean without 1000 preprocessor #if-statements

render/fx_renderer/tracy.c
```diff
@@ -0,0 +1,360 @@
...
+		.context = tracy_data->context_id,
+		.queryId = (uint16_t) query,
+	};
+	___tracy_emit_gpu_zone_end(data);
+}
+
+void tracy_gpu_context_collect(struct tracy_data *tracy_data) {
+	if (tracy_data == NULL) {
+		// TODO: fix compilation error
+		// TracyCMessageL("tracy_data == NULL");
```
WillPower3309: TODO

ErikReider:
> instead of `HEAD` the wrap should probably pin `tracy` to releases, so API doesn't break randomly

Fixed by [aa2c59c](https://github.com/wlrfx/scenefx/pull/129/commits/aa2c59c39e398d2b4b0642dabeb92e628cb49645)

render/fx_renderer/tracy.c
```diff
@@ -0,0 +1,352 @@
...
+	GLint draw_fbo_id = 0, read_fbo_id = 0;
+	glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &draw_fbo_id);
+	glGetIntegerv(GL_READ_FRAMEBUFFER_BINDING, &read_fbo_id);
+
+	// From the tracy documentation:
+	//	https://github.com/wolfpld/tracy/releases/latest/download/tracy.pdf
+	while (data->queue.size > 0) {
+		int *queue_data = data->queue.data;
+		const int fence_i = queue_data[0];
+		if (glClientWaitSync(data->fence[fence_i], 0, 0) == GL_TIMEOUT_EXPIRED) {
```
the-eater: queue basically breaks if one fence is expired, i forgot how I fixed it, but I had to fix it when I was testing my blur performance fixes on old hardware

ErikReider: LMAO, I was just looking at this segfault 🤣

ErikReider: Fixed by [4a853c4](https://github.com/wlrfx/scenefx/pull/129/commits/4a853c44f7b8699172eaad84a6f291f197613602)

the-eater: you should still clean up the expired fence and advance the queue, otherwise you'll run out of fences

ErikReider: > you should still clean up the expired fence and advance the queue, otherwise you'll run out of fences

ca97ea03

render/fx_renderer/fx_pass.c
```diff
@@ -15,8 +15,9 @@
 #include "render/fx_renderer/fx_renderer.h"
 #include "render/fx_renderer/shaders.h"
 #include "render/pass.h"
-#include "scenefx/render/fx_renderer/fx_renderer.h"
+#include "render/tracy.h"
 #include "scenefx/render/fx_renderer/fx_effect_framebuffers.h"
+#include "scenefx/render/fx_renderer/fx_renderer.h"
```
WillPower3309, line 20: oop nice catch

types/scene/wlr_scene.c
WillPower3309, line 1: hrm too bad these can't be moved to fx_pass. I don't like having it outside of the renderer

WillPower3309 (approved):

types/scene/wlr_scene.c
ErikReider: I agree. The regular ZONE calls are only for profiling the times spent on the CPU, so we could remove those. The `TRACY_MARK_FRAME` could be moved into the passes submit function, but it would be called for other render passes as well, like within `wlr_ext_image_copy_capture_v1`.

What about only keeping the `TRACY_MARK_FRAME`, removing the CPU profiling until we find a better solution?

WillPower3309: Hmm, yeah I think I agree with this. Let's keep it out of scene for now

Issue #108, opened 2025-02-23
Move ignore transparent functionality to its own shader


WillPower3309:
For micro performance gains in our tex shader

PR #133, opened 2025-07-13
ignore transparent improvements

### LINKED ISSUE #132: blur_ignore_transparent cutout incorrect on rotated displays
_Obligatory: I'm not sure if this is an issue with SwayNC, SwayFX, or SceneFX. While I've only seen it happen with SwayNC, based on https://github.com/ErikReider/SwayNotificationCenter/issues/572, this sounds like it may be an issue with SwayFX or SceneFX_

**Describe the bug**
As tested with swayfx 0.5 and 0.5.3 (and thus corresponding scenefx versions), using the following config to get a background blur on SwayNC leads to the blur behind swaync's control center and notifications no longer matching the window's location (see image below) for displays in portrait mode.

```
layer_effects "swaync-notification-window" {
    blur enable;
    blur_ignore_transparent true;
  }
layer_effects "swaync-control-center" {
    blur enable;
    blur_ignore_transparent true;
}
```

Based on my testing, this seems to be an issue strictly with the `blur_ignore_transparent` setting: when the setting is disabled, blurs are correctly positioned behind their application windows. 

**To Reproduce**
1. Use a window manager such as swayfx to blur the background of swaync
2. Rotate your display into portrait mode and open swaync (note: I've tried launching swaync only after the screen has been rotated, and the problem still exists) 
3. The blurred portion of the screen will mot match swaync's location on the screen

**Expected behavior**
For the blurred area behind the swaync windows to match the window's extents. 

**Please provide logs if you're experiencing notification errors / bugs**
N/A

**Screenshots**

Image

WillPower3309:
Will close #132  and #108

### DIFF
--- a/render/fx_renderer/fx_pass.c
+++ b/render/fx_renderer/fx_pass.c
@@ -16,6 +16,7 @@
 #include "render/pass.h"
 #include "scenefx/render/fx_renderer/fx_renderer.h"
 #include "scenefx/render/fx_renderer/fx_effect_framebuffers.h"
+#include "scenefx/render/pass.h"
 #include "scenefx/types/fx/corner_location.h"
 #include "scenefx/types/fx/blur_data.h"
 #include "util/matrix.h"
@@ -27,7 +28,6 @@ struct fx_render_texture_options fx_render_texture_options_default(
 	struct fx_render_texture_options options = {
 		.corner_radius = 0,
 		.corners = CORNER_LOCATION_NONE,
-		.discard_transparent = false,
 		.clip_box = NULL,
 	};
 	memcpy(&options.base, base, sizeof(*base));
@@ -341,8 +341,7 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 
 	bool has_alpha = texture->has_alpha
 		|| alpha < 1.0
-		|| fx_options->corner_radius > 0
-		|| fx_options->discard_transparent;
+		|| fx_options->corner_radius > 0;
 	setup_blending(!has_alpha ? WLR_RENDER_BLEND_MODE_NONE : options->blend_mode);
 
 	glUseProgram(shader->program);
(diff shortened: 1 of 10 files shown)


render/fx_renderer/fx_pass.c
```diff
@@ -945,14 +1027,13 @@ void fx_render_pass_add_blur(struct fx_gles_render_pass *pass,
...
 	// Get a stencil of the window ignoring transparent regions
-	if (fx_options->ignore_transparent && fx_options->tex_options.base.texture) {
+	if (should_ignore_transparent) {
 		stencil_mask_init();
-
-		struct fx_render_texture_options tex_options = fx_options->tex_options;
-		tex_options.discard_transparent = true;
-		fx_render_pass_add_texture(pass, &tex_options);
-
+		fx_render_pass_add_discard_transparent(pass, &tex_options->base);
```
ErikReider, line 1036: Wouldn't it be better to just create a `fx_render_pass_add_blur_texture` or something which does what these lines + lines 959->969? Then we should be able to remove the stenciling right?

include/render/fx_renderer/fx_renderer.h
```diff
@@ -186,6 +186,9 @@ struct fx_renderer {
 		struct blur_shader blur1;
 		struct blur_shader blur2;
 		struct blur_effects_shader blur_effects;
+		struct discard_transparent_shader discard_transparent_rgba;
+		struct discard_transparent_shader discard_transparent_rgbx;
```
WillPower3309, line 190: RGBX shouldn't be needed, since it has no opacity

render/fx_renderer/gles3/shaders/discard_transparent.frag
```diff
@@ -0,0 +1,40 @@
...
+	return texture2D(tex, v_texcoord);
+#elif SOURCE == SOURCE_TEXTURE_RGBX
+	return vec4(texture2D(tex, v_texcoord).rgb, 1.0);
+#endif
+}
+
+void main() {
+	if (sample_texture().a == 0.0) {
+		discard;
+		return;
```
WillPower3309, line 38: Not needed

PR #135, opened 2025-07-24
Add drop shadows (CSS drop-shadows)

ErikReider:
Allows for a dynamic shadow that isn't a fixed shape, like our shadow_node. This allows for shadows that "exclude" transparent regions like weirdly shaped surfaces and text buffers.

So according to the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow), the drop-shadow works by applying a blur to the base textures alpha mask. So this should mimic the implementation that's used by web browsers. So a 10px radii in the browser should result in the same looking blur here.

image

Swaync with the smart shadows enabled:
image

TODO:
- [x] Change name to "drop-shadow"? (from smart shadows)
- [ ] ~Use kawase blur instead?~ (Use a two-pass Gaussian blur, like the web browsers use)
- [x] Add this to the existing shadow-node and allow it to link to a `scene_buffer` instead?
- [x] Fix damage not being compensated for. Needs the same treatment as regular blur, so maybe do both in the same pass?
- [x] Fix weird weakness of color with a "low" alpha like 0.5
- [x] Add all parameters
  - [x] x, y (also compensate the node visible region with these parameters)
  - [x] Fix node visible region by creating two regions (one for blur and one for shadow) and run the union pixman operation
  - [x] Color (pre-multiplied?)
- [ ] ~Add post-processing or use the regular `pass_add_texture` function instead of a dedicated shader (or just blit for gles3?)~
- [x] Add gles2 support
- [x] Make sure scaling and transforms work as expected
- [x] Fix all commented TODOs
- [ ] Rebase + eventual fixes
  - [ ] Fix blur damage breaking with the latest changes in main
- [ ] Ensure optimal performance


render/fx_renderer/fx_pass.c
```diff
@@ -1015,6 +1026,219 @@ bool fx_render_pass_add_optimized_blur(struct fx_gles_render_pass *pass,
 	return !failed;
 }
 
+static void render_drop_shadow_blur_direction(struct fx_gles_render_pass *pass,
+		const struct fx_render_drop_shadow_options *fx_options,
+		bool is_horizontal) {
+	const struct wlr_render_texture_options *options = &fx_options->tex_options.base;
+	struct fx_renderer *renderer = pass->buffer->renderer;
+	struct fx_texture *texture = fx_get_texture(options->texture);
+	assert(texture->has_alpha);
```
WillPower3309, line 982: are we sure we want to make this assertion? This would put the onus on the compositor to check if the texture has alpha. In the case of a scene buffer without alpha, should we fall back to the box shadow instead?

render/fx_renderer/fx_pass.c
```diff
@@ -1015,6 +1026,219 @@ bool fx_render_pass_add_optimized_blur(struct fx_gles_render_pass *pass,
 	return !failed;
 }
 
+static void render_drop_shadow_blur_direction(struct fx_gles_render_pass *pass,
+		const struct fx_render_drop_shadow_options *fx_options,
```
WillPower3309: I'm wondering if instead we just pass the color, base options, and sigma since that's all we use.

render/fx_renderer/fx_pass.c
```diff
@@ -1015,6 +1026,219 @@ bool fx_render_pass_add_optimized_blur(struct fx_gles_render_pass *pass,
...
+
+	glBindTexture(texture->target, 0);
+
+	glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
+
+	pop_fx_debug(renderer);
+}
+
+static void fx_render_pass_add_drop_shadow_final(struct fx_gles_render_pass *pass,
+		const struct fx_render_drop_shadow_options *fx_options) {
```
WillPower3309: I'm wondering if instead we just pass the color and base options, since that's all we use.

types/scene/wlr_scene.c
```diff
@@ -41,11 +41,11 @@
 #define HIGHLIGHT_DAMAGE_FADEOUT_TIME   250
 
 #define SCENE_BUFFER_SHOULD_BLUR(scene_buffer, blur_data) \
-	scene_buffer->backdrop_blur && is_scene_blur_enabled(blur_data) && \
-		(!scene_buffer->buffer_is_opaque || scene_buffer->opacity < 1.0f)
+	(scene_buffer->backdrop_blur && is_scene_blur_enabled(blur_data) && \
+	 (!scene_buffer->buffer_is_opaque || scene_buffer->opacity < 1.0f))
```
WillPower3309: nit: I'd rather habe the tab than a space here

types/scene/wlr_scene.c
```diff
@@ -41,11 +41,11 @@
...
 #define SCENE_RECT_SHOULD_BLUR(scene_rect, blur_data) \
-	scene_rect->backdrop_blur && is_scene_blur_enabled(blur_data) && \
-		scene_rect->color[3] < 1.0f
+	(scene_rect->backdrop_blur && is_scene_blur_enabled(blur_data) && \
+	 scene_rect->color[3] < 1.0f)
```
WillPower3309: same as above

types/scene/wlr_scene.c
```diff
@@ -1010,7 +1014,7 @@ void wlr_scene_shadow_set_corner_radius(struct wlr_scene_shadow *shadow, int cor
 }
 
 void wlr_scene_shadow_set_blur_sigma(struct wlr_scene_shadow *shadow, float blur_sigma) {
-	if (shadow->blur_sigma == blur_sigma) {
+	if (shadow->blur_sigma == blur_sigma || blur_sigma < 0) {
```
WillPower3309: an aside, but I think we should try to figure out how to handle out of bounds values more formally. I think upstream tends to make an assertion, but we haven't really formed an opinion yet. Curious to hear your thoughts - I think we should assert to match upstream

WillPower3309 (changes requested):

WillPower3309:
Still need to look at wlr_scene.c

render/fx_renderer/fx_pass.c
ErikReider: This asserts that our internal effect buffer textures (effects_buffer and effects_buffer_swapped) supports alpha, which always should be the case. Would you prefer that I remove the assertion?

types/scene/wlr_scene.c
ErikReider: Yeah, lets match upstream! :D

include/scenefx/types/fx/blur_data.h
```diff
@@ -4,6 +4,9 @@
 #include <stdbool.h>
 #include <wlr/util/addon.h>
 
+#define drop_shadow_calc_size(blur_sigma) (ceil(1.5f * blur_sigma) * 2)
```
WillPower3309, line 8: Q: does an equal blur radius value result in an equal spread in drop shadow and box shadow?

include/scenefx/types/wlr_scene.h
```diff
@@ -156,6 +157,11 @@ struct wlr_scene_rect {
 	struct clipped_region clipped_region;
 };
 
+enum wlr_scene_shadow_type {
+	WLR_SCENE_SHADOW_TYPE_DROP,
+	WLR_SCENE_SHADOW_TYPE_BOX,
+};
```
WillPower3309: do we need these? We should be able to tell if buffer_link is NULL or not

include/scenefx/types/wlr_scene.h
```diff
@@ -561,6 +577,40 @@ void wlr_scene_shadow_set_blur_sigma(struct wlr_scene_shadow *shadow, float blur
...
+ * NOTE: Will fallback to a box-shadow if the provided buffer is NULL.
+ */
+void wlr_scene_shadow_set_reference_buffer(struct wlr_scene_shadow *shadow,
+		struct wlr_scene_buffer *ref_buffer);
+
+/**
+ * Gets the sample size of the shadow. Can be used when setting the size of
+ * the shadow node. Returns the correct size offset used depending on the
+ * shadow-type. A drop-shadow returns a larger integer compared to the
+ * box-shadow, so this can be used to dynamically adjust the size and position
```
WillPower3309: why would this be?

include/scenefx/types/wlr_scene.h
```diff
@@ -561,6 +577,40 @@ void wlr_scene_shadow_set_blur_sigma(struct wlr_scene_shadow *shadow, float blur
...
+ * the buffer (a drop-shadow, like in CSS). This type of shadow adapts to the
+ * buffer, ignoring transparent regions. An example would be text shadows, or
+ * weirdly shaped bars/docks. NOTE: Will ignore the clipped_region.
+ *
+ * Otherwise, a regular faster and simpler box-shadow will be drawn.
+ *
+ * NOTE: Please use the `wlr_scene_shadow_get_offset` function to get a dynamic
+ * offset depending on the type.
+ */
+void wlr_scene_shadow_set_type(struct wlr_scene_shadow *shadow, enum wlr_scene_shadow_type type);
```
WillPower3309: I think we can remove this call for now, if we're aligned on my earlier comment

include/types/linked_nodes.h
```diff
@@ -0,0 +1,27 @@
+#ifndef TYPES_LINKED_NODES_H
```
WillPower3309, line 1: nit: this file should be `linked_node.h`

types/scene/linked_nodes.c
```diff
@@ -0,0 +1,78 @@
+#include <assert.h>
```
WillPower3309, line 1: nit: file should be `linked_node.c`

types/scene/linked_node.c
```diff
@@ -0,0 +1,78 @@
...
+static bool linked_nodes_are_linked(struct linked_node *node_1,
+		struct linked_node *node_2) {
+	return node_1->link != NULL && node_2->link != NULL
+		&& node_1->link == node_2->link;
+}
+
+void linked_node_init_link(struct linked_node *node_1,
+		struct linked_node *node_2) {
+	if (node_1->link || node_2->link) {
+		assert(linked_nodes_are_linked(node_1, node_2));
```
WillPower3309: I think it'd be enough to assert `node_1->link == node_2->link` since we already check that one of them is not NULL

render/fx_renderer/gles2/shaders/drop_shadow_final.frag
```diff
@@ -0,0 +1,19 @@
...
+precision mediump float;
+#endif
+
+varying vec2 v_texcoord;
+uniform sampler2D tex;
+
+uniform vec2 size;
+uniform float blur_sigma;
+uniform vec4 color;
+uniform bool is_horizontal;
```
WillPower3309: size, blur_sigma and is_horizontal are unused

render/fx_renderer/gles3/shaders/drop_shadow_final.frag
```diff
@@ -0,0 +1,19 @@
...
+
+precision highp float;
+
+in vec2 v_texcoord;
+uniform sampler2D tex;
+
+uniform vec2 size;
+uniform float blur_sigma;
+uniform vec4 color;
+uniform bool is_horizontal;
```
WillPower3309: size, blur_sigma, is_horizontal unused

types/scene/wlr_scene.c
```diff
@@ -1027,6 +1041,37 @@ void wlr_scene_shadow_set_color(struct wlr_scene_shadow *shadow, const float col
...
+}
+
+int wlr_scene_shadow_get_offset(struct wlr_scene_shadow *shadow) {
+	switch (shadow->type) {
+	case WLR_SCENE_SHADOW_TYPE_DROP:
+		return drop_shadow_calc_size(shadow->blur_sigma);
+	default:
+		return shadow->blur_sigma;
+	}
+}
```
WillPower3309: have you tested that the shadows are functionally identical between box and drop? We'd want the same appearance minus the transparent areas between types. Also, why do you refer to the sigma as offset for the drop shadow?

types/scene/wlr_scene.c
```diff
@@ -1354,6 +1400,31 @@ void wlr_scene_buffer_set_buffer_with_options(struct wlr_scene_buffer *scene_buf
 		pixman_region32_intersect(&output_damage, &output_damage, &cull_region);
 		pixman_region32_fini(&cull_region);
 
+		// Also damage the linked drop-shadow
+		struct linked_node *shadow_link = linked_nodes_get_sibling(&scene_buffer->drop_shadow_link);
+		if (shadow_link != NULL) {
+			struct wlr_scene_shadow *scene_shadow = wl_container_of(shadow_link, scene_shadow, buffer_link);
+			if (scene_shadow && SCENE_SHADOW_IS_DROP(scene_shadow)) {
```
WillPower3309: do we need this check? We already have an assertion on the link. Maybe we should add another assertion here instead

types/scene/wlr_scene.c
```diff
@@ -1960,6 +2032,92 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+
+			struct wlr_texture *texture = scene_buffer_get_texture(scene_buffer,
+					data->output->output->renderer);
+			if (texture == NULL) {
+				goto fallback_box_shadow;
+			}
+
+			const int shadow_size = drop_shadow_calc_size(scene_shadow->blur_sigma) * data->scale;
+
+			// Optimization: Fallback to a regular box-shadow if the texture is fully opaque
```
WillPower3309: beautiful

types/scene/wlr_scene.c
WillPower3309: Realize I already left this comment earlier in blur_data.h

include/scenefx/types/fx/blur_data.h
ErikReider: Yes, but the overall color alpha will vary depending on the textures opacity (should we fix this?)

image

image

WillPower3309: Hmm, ideally they'd look the same but I haven't dived deep enough to know if this is a trivial problem to solve or not haha

WillPower3309: Unfortunately the single pixel transparent area between the shadow and texture would need texture shader side changes (so we can keep that out of scope)

ErikReider: > Hmm, ideally they'd look the same

Yeah, I noticed that the drop shadow spread is a bit further than the box one, so I'll be working on fixing that after the university finals season has passed

ErikReider:
Converting to draft as I think that I can optimize the shader a bit more like the blur shader (maybe even use that shader instead?)

PR #138, opened 2025-10-04
blur: add clipping region

the-eater:
Currently blur does not respect the clipping region

So when blur is enabled on an element, and a clipping region is applied, the clipping region is filled with blur.

My [PR](https://github.com/WillPower3309/swayfx/pull/457) on swayfx is affected by this as shown by discord user ahf

TEMP

Both the top and bottom parts shouldn't be covered by blur and simply show the contents of the main buffer, however the border is rendered after the content, and overlays it with the blur.

Adding this clipping code fixes it.

### DIFF
--- a/render/fx_renderer/fx_pass.c
+++ b/render/fx_renderer/fx_pass.c
@@ -347,10 +347,29 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 		|| fx_options->discard_transparent;
 	setup_blending(!has_alpha ? WLR_RENDER_BLEND_MODE_NONE : options->blend_mode);
 
+	pixman_region32_t clip_region;
+	if (options->clip) {
+		pixman_region32_init(&clip_region);
+		pixman_region32_copy(&clip_region, options->clip);
+	} else {
+		pixman_region32_init_rect(&clip_region, dst_box.x, dst_box.y, dst_box.width, dst_box.height);
+	}
 	const struct wlr_box clipped_region_box = fx_options->clipped_region.area;
 	enum corner_location clipped_region_corners = fx_options->clipped_region.corners;
 	int clipped_region_corner_radius = clipped_region_corners != CORNER_LOCATION_NONE ?
 		fx_options->clipped_region.corner_radius : 0;
+	if (!wlr_box_empty(&clipped_region_box)) {
+		pixman_region32_t user_clip_region;
+		pixman_region32_init_rect(
+			&user_clip_region,
+			clipped_region_box.x + clipped_region_corner_radius * 0.3,
+			clipped_region_box.y + clipped_region_corner_radius * 0.3,
+			fmax(clipped_region_box.width - clipped_region_corner_radius * 0.6, 0),
+			fmax(clipped_region_box.height - clipped_region_corner_radius * 0.6, 0)
+		);
+		pixman_region32_subtract(&clip_region, &clip_region, &user_clip_region);
+		pixman_region32_fini(&user_clip_region);
+	}
 
 	glUseProgram(shader->program);
 
@@ -398,7 +417,8 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 	set_proj_matrix(shader->proj, pass->projection_matrix, &dst_box);
 	set_tex_matrix(shader->tex_proj, options->transform, &src_fbox);
 
-	render(&dst_box, options->clip, shader->pos_attrib);
+	render(&dst_box, &clip_region, shader->pos_attrib);
+	pixman_region32_fini(&clip_region);
 
 	glBindTexture(texture->target, 0);
 	pop_fx_debug(renderer);


the-eater:
Both of these are made with the current head in my branch in swayfx, no changes will be needed there

With unpatched scenefx:

20251006_15h47m03s_grim

With this patch

20251006_15h47m31s_grim

ErikReider:
> Both of these are made with the current head in my branch in swayfx, no changes will be needed there

Are you applying the blur to the border scene_buffer? I'm not really getting that's being shown here

the-eater:
@ErikReider I'm applying blur to the scene rect's of the border's and titlebar, as seen here

https://github.com/the-eater/swayfx/commit/7eed52a8f59d5b9481a36508de8e963e95fc7b73

The top and bottom border in screenshots are made taller to accommodate the border radius, and then a clipping region is used to still display the content behind it.

e.g. without the patch but also without blur it displays fine

20251006_20h17m51s_grim

render/fx_renderer/fx_pass.c
```diff
@@ -361,6 +371,8 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 		break;
 	}
 
+
+
```
ErikReider: Nit: Remove the empty lines

types/scene/wlr_scene.c
```diff
@@ -1838,6 +1838,18 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 
 		corner_location_transform(node_transform, &rect_corners);
 
+
```
ErikReider: Nit: remove extra empty line

types/scene/wlr_scene.c
```diff
@@ -1872,23 +1884,18 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 				.use_optimized_blur = scene_rect->backdrop_blur_optimized,
 				.blur_data = &scene->blur_data,
 				.ignore_transparent = false,
+				.clipped_region = {
+					.area = rect_clipped_region_box,
+					.corner_radius = rect_clipped_region_corner_radius * data->scale,
+					.corners = rect_clipped_corners,
+				}
```
ErikReider: Don't forget about adding clipped blur to the `scene_buffer` switch case as well

types/scene/wlr_scene.c
```diff
@@ -1872,23 +1884,18 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 				.use_optimized_blur = scene_rect->backdrop_blur_optimized,
 				.blur_data = &scene->blur_data,
 				.ignore_transparent = false,
+				.clipped_region = {
```
ErikReider, line 1925: Also, don't forget to initialize the `wlr_render_texture_options` clipped_region to `{0}` everywhere.

ErikReider (changes requested):
Otherwise, LGTM

types/scene/wlr_scene.c
the-eater: does `{0}` "just" initialize all the fields to zero? I don't have to define every field?

ErikReider: Yes. It's a lot easier :)

the-eater: Wait do you mean `fx_render_texture_options`? because I didn't change `wlr_render_texture_options`

ErikReider: > Wait do you mean `fx_render_texture_options`? because I didn't change `wlr_render_texture_options`

Sorry, I meant `fx_render_texture_options`. We always want it to be set to 0, instead of letting the compiler choose.

types/scene/wlr_scene.c
ErikReider: Nvm, ignore this one

types/scene/wlr_scene.c
the-eater: Done :)

render/fx_renderer/fx_pass.c
```diff
@@ -345,6 +346,11 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 		|| fx_options->discard_transparent;
 	setup_blending(!has_alpha ? WLR_RENDER_BLEND_MODE_NONE : options->blend_mode);
 
+	const struct wlr_box clipped_region_box = fx_options->clipped_region.area;
+	enum corner_location clipped_region_corners = fx_options->clipped_region.corners;
+	int clipped_region_corner_radius = clipped_region_corners != CORNER_LOCATION_NONE ?
+		fx_options->clipped_region.corner_radius : 0;
+
```
WillPower3309, line 373: I'd like to optimize things a bit by subtracting the clipped region box from the damaged area like we do with rect and shadow

the-eater: Ah! makes sense, will do :)

the-eater: I think the new changes do what you're asking for

the-eater:
clipping in favor of #149

PR #139, opened 2025-10-06
Fixes segfault on GPU reset

ErikReider:
The compositor would destroy the fx-renderer, destroying all fx_framebuffers, but the outputs fx_effect_framebuffers wouldn't be affected. This would lead to a use after free.

Found this when Valheim would cause RADV to reset the GPU xD

### DIFF
--- a/render/fx_renderer/fx_effect_framebuffers.c
+++ b/render/fx_renderer/fx_effect_framebuffers.c
@@ -26,8 +26,6 @@ static void addon_handle_destroy(struct wlr_addon *addon) {
 	destroy_fx_framebuffer(&fbos->effects_buffer);
 	destroy_fx_framebuffer(&fbos->effects_buffer_swapped);
 
-	pixman_region32_fini(&fbos->blur_padding_region);
-
 	free(fbos);
 }
 


render/fx_renderer/fx_framebuffer.c
```diff
@@ -161,6 +162,12 @@ void fx_framebuffer_bind(struct fx_framebuffer *fx_buffer) {
 }
 
 void fx_framebuffer_destroy(struct fx_framebuffer *fx_buffer) {
+	if (!fx_buffer->initialized) {
```
WillPower3309: Q: can we replace this with a null check?

ErikReider: Fixed :)

WillPower3309: I meant the initialized variable in general, should we instead check if the buffer is null? I might be missing something but it looks like it'd be null when we expect the initialized to be false

ErikReider: Good catch! Fixed :)

WillPower3309 (approved):
LGTM

PR #144, opened 2025-10-16
fix black being transparent

### LINKED ISSUE #448: Setting pure black on titlebars makes things disappear
### Please fill out the following:
**Swayfx Version:** `swayfx version 0.5 (based on sway 1.10.1)`

**Debug log:** https://gist.github.com/cccslater/e331e3fb6c7818bff40c5b862a5e6839#file-log

**Configuration File:** https://gist.github.com/cccslater/e331e3fb6c7818bff40c5b862a5e6839#file-config
  - I believe this would happen with any config file
  - I haven't tried with the default because I've had a long week, if you can't reproduce let me know and I'll stop being lazy!

**Description:**
  - Enable titlebar
  - `swaymsg client.[focused, focused_inactive, etc.] 000000 000000 ffffff 000000 000000`
  - Observe that, when [focused, focused + inactive, etc.], most of the titlebar elements disappear
  - Despair

Image

I imagine this goes upstream but haven't checked yet.

the-eater:
This fixes https://github.com/WillPower3309/swayfx/issues/448

Basically the original code expects the background to just, Be Black and thus does not render that area -however- since scenefx supports transparency it just means that it renders over it and pure black rects and single pixel buffers are hidden

Before:

20251016_13h11m24s_grim

After:

20251016_13h10m02s_grim

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -2411,15 +2411,6 @@ struct render_list_constructor_data {
 	bool fractional_scale;
 };
 
-static bool scene_buffer_is_black_opaque(struct wlr_scene_buffer *scene_buffer) {
-	return scene_buffer->is_single_pixel_buffer &&
-		scene_buffer->single_pixel_buffer_color[0] == 0 &&
-		scene_buffer->single_pixel_buffer_color[1] == 0 &&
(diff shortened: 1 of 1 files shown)


ErikReider:
Just adding a note here: we should check if this is a bug in upstream or not

PR #145, opened 2025-10-19
new node: blur source

the-eater:
Currently the blur applied to nodes is Locally Sourced Free Range Blur. which is fine if there's e.g. only 1 thing in a vacuum needing to be blurred. but when other things need to be blurred too that are rendered next to it, it causes the issue that the blur leaks into each other, object A will first blur, then object B will sample the existing rendered frame for its own blur.

This causes mostly positional darkness on most blurred backgrounds.

The solution provided in this PR, is to have a dedicated node that gets rendered before -anything- else, which will then pre-render the blur to a texture

This node is called "blur source" as, it will be the source of blur for other nodes

A node (limited to rect and buffer atm, as they are only receiving ends of blur) can be added as a target to the blur, this will establish a link between the node and the blur source, allowing us to limit the region we blur / disable the blur dynamically if no nodes are actually doing blurring

An example of the results with my swayfx PR

Before|After|Difference
---|---|---
20251020_16h41m52s_grim | 20251020_16h44m28s_grim | Diff

This PR doesn't change old blurring code, it merely splits the code apart so a texture can be fed to the blur pass and reused.

Performance should be the exact same for everything, (with exception of my PR which should see minimal improved performance because blurring is only done once instead of being ran, 5-8 times per container)

## Todo:

- [x] Reduce region to render blur based on listed and enabled blur
- [x] Add flags to disable/enable blur for node
- [x] Detect when blur is outside of blur source region, and let it fallback to local blur

## Credits

`linked_node.c/h` is from #135 by @ErikReider, I simply added the `linked_node_list`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ wlroots is the de-facto library for building wayland compositors, and its scene 
 
 **Please note: while SceneFX is in use by SwayFX version 0.4, it is not yet ready for usage by other compositors. Please refer to the [1.0 milestone](https://github.com/wlrfx/scenefx/milestone/2) to track the remaining tasks for our stable 1.0 release**
 
+## Compositors Using SceneFX
+Plenty of popular wayland compositors are using SceneFX to render eyecandy, including:
+- [SwayFX](https://github.com/WillPower3309/swayfx)
+- [MangoWC](https://github.com/DreamMaoMao/mangowc)
+- [mwc](https://github.com/nikoloc/mwc)
+- dwl [with a patch](https://codeberg.org/dwl/dwl-patches/src/branch/main/patches/scenefx)
+
 ## Installation
 <a href="https://repology.org/project/scenefx/versions"><img src="https://repology.org/badge/vertical-allrepos/scenefx.svg"/></a>
 


the-eater:
Reducing region is, unclear as I need to know the absolute position of nodes :thinking:

WillPower3309:
Looks like there are some conflicts, likely from the blur strength PR getting merged

include/scenefx/render/pass.h
```diff
@@ -97,14 +97,20 @@ struct fx_render_box_shadow_options {
...
 	bool ignore_transparent;
 	float blur_strength;
 };
 
+struct fx_render_apply_blur_pass_options {
+	struct fx_render_texture_options tex_options;
+	pixman_region32_t *opaque_region;
+	bool ignore_transparent;
+	struct wlr_texture *blur_source;
+};
```
WillPower3309: What's the reasoning behind two different structs?

WillPower3309 (commented):
I might not have communicated my previous comment well, but I think that right now this is going about blur nodes incorrectly. Instead of having a `blur_source`, we should completely remove blurring as part of the rect and buffer node rendering logic and create a blur node. Compositors can then move / clip / do whatever they like to this node. The blur node would just render a blurred texture of whatever is beneath it. We can move the `backdrop_blur_optimized` logic to it as well in order to just render the blur from the bottom layer. This would keep the optimized blur node, and would just be the optimized logic from the rect and buffer node rendering.

The transparent discarding could be hard to do correctly (and may be the one use case for requiring a `linked_node`), but if it causes any trouble, I'd rather you just remove it from the buffer and rect nodes along with the rest of the blur logic. I could then go ahead and add it back onto this branch as part of your blur node.

Ideally we'd have a completely decoupled blur node, and all blur logic purged from rect and buffer nodes. In fact, we shouldn't need the linked_node struct until we add back that transparent discarding. I'd suspect we should end up with more removed code than added code with this approach :)

the-eater:
I see! that makes more sense, I think i really went all in on the concept that was stuck in my head! Time to try again, this time will be faster lmao.

The only worry I have is having 1 clipping rect not being adequate. and having a `pixman_region` as the render area might make things easier to selectively render?

include/scenefx/render/pass.h
the-eater: because they're 2 different passes, that needed different inputs, and this made it clearer which pass needed which.

WillPower3309:
> I see! that makes more sense, I think i really went all in on the concept that was stuck in my head! Time to try again, this time will be faster lmao.
> 
> The only worry I have is having 1 clipping rect not being adequate. and having a `pixman_region` as the render area might make things easier to selectively render?

we really should revamp our clipping logic at some point, I like the idea of having a region instead of a box. For now though the box should be adequate. Either way this would likely replace #138

PR #146, opened 2025-10-22
add scene_buffer_crossfade struct and enum

WillPower3309:


WillPower3309:
@the-eater if you have time and wouldn't mind giving this a review I'd appreciate a once over!

render/fx_renderer/gles3/shaders/tex_crossfade.frag
```diff
@@ -0,0 +1,71 @@
...
+#endif
+
+precision highp float;
+
+in vec2 v_texcoord;
+
+#if SOURCE == SOURCE_TEXTURE_EXTERNAL
+uniform samplerExternalOES tex_prev;
+uniform samplerExternalOES tex_next;
+#elif SOURCE == SOURCE_TEXTURE_RGBA || SOURCE == SOURCE_TEXTURE_RGBX
```
the-eater, line 24: so just so i am getting this correctly, this is basically `if nvidia: X else: Y`? 

So it's unlikely to get a source External and a dest RGBX?

render/fx_renderer/fx_pass.c
```diff
@@ -343,8 +343,7 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
 
 	bool has_alpha = texture->has_alpha
 		|| alpha < 1.0
-		|| fx_options->corner_radius > 0
-		|| fx_options->discard_transparent;
```
the-eater, line 368: this an accidental removal or on purpose?

types/scene/wlr_scene.c
```diff
@@ -2042,6 +2176,7 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 		enum wl_output_transform transform =
 			wlr_output_transform_invert(scene_buffer->transform);
 		transform = wlr_output_transform_compose(transform, data->transform);
+		enum corner_location buffer_corners = scene_buffer->corners;
```
the-eater: Nit: changes like these probably can best put in their own PR/commit (depending on if you're planning to squash and rebase or just, rebase, the latter is easiest)

types/scene/wlr_scene.c
```diff
@@ -2080,6 +2215,51 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 			});
 		}
 		break;
+
+	case WLR_SCENE_NODE_BUFFER_CROSSFADE:;
+		struct wlr_scene_buffer_crossfade *scene_buffer_crossfade = wlr_scene_buffer_crossfade_from_node(node);
+
+		struct wlr_scene_buffer *scene_buffer_prev = scene_buffer_crossfade->scene_buffer_prev;
+		struct wlr_scene_buffer *scene_buffer_next = scene_buffer_crossfade->scene_buffer_next;
+		if (scene_buffer_prev == NULL || scene_buffer_next == NULL) {
```
the-eater, line 2239: is `prev` or `next` allowed to be `NULL` when crossfade is either at `0` or at `1`?

Possible optimization of falling back to `add_texture` in that case.

(unsure what real world usage looks like)

render/fx_renderer/gles3/shaders/tex_crossfade.frag
```diff
@@ -0,0 +1,71 @@
...
+#elif SOURCE == SOURCE_TEXTURE_RGBX
+	return vec4(texture2D(tex_prev, v_texcoord).rgb, 1.0);
+#endif
+}
+
+vec4 sample_texture_next() {
+#if SOURCE == SOURCE_TEXTURE_RGBA || SOURCE == SOURCE_TEXTURE_EXTERNAL
+	return texture2D(tex_next, v_texcoord);
+#elif SOURCE == SOURCE_TEXTURE_RGBX
+	return vec4(texture2D(tex_next, v_texcoord).rgb, 1.0);
```
the-eater, line 53: how likely is it to get a `RGBX` to/from `RGBA` texture crossfaded?

render/fx_renderer/fx_pass.c
```diff
@@ -424,6 +423,132 @@ void fx_render_pass_add_texture(struct fx_gles_render_pass *pass,
...
+		abort();
+	}
+
+	// TODO: wlr_render_texture_options_get_src_box / wlr_render_texture_options_get_dst_box
+	struct wlr_box dst_box = options->dst_box;
+	struct wlr_fbox src_fbox = options->src_box;
+	// TODO: wlr_render_texture_options_get_alpha
+	float alpha = options->alpha != NULL ? *options->alpha : 1;
+
+	// TODO: take texture_next into consideration?
```
the-eater, line 453: they might need different src_box/destination boxes yes, somewhere in my many branches i have code to make a second `tex_proj` based on the projection matrix of the first texture

the-eater (commented):
unable to test since the `swayfx` branch seems out of date, but here you go, a quick lookover

render/fx_renderer/fx_pass.c
the-eater: Here's the code I used https://github.com/the-eater/scenefx/blob/blur-source-mask/render/fx_renderer/fx_pass.c#L447-L473

render/fx_renderer/gles3/shaders/tex_crossfade.frag
WillPower3309: The intention is for it to be used against two textures of the same type, and we make an assertion for that as well

render/fx_renderer/fx_pass.c
WillPower3309: Actually intentional; maybe best suited for a different PR, but the discarding of transparent pixels wouldn't make the texture itself transparent (it would already need to have transparent elements) so removing that line fixes a logical error

types/scene/wlr_scene.c
WillPower3309: Yeah good idea! I actually had a similar thought, looks like I forgot to add that here

render/fx_renderer/gles3/shaders/tex_crossfade.frag
WillPower3309: Should be impossible given our assertions

render/fx_renderer/fx_pass.c
WillPower3309: Thanks for the link, I actually want them to be overlaid and stretched over each other so we should be able to reuse the same boxes, helps with the resize animation

WillPower3309:
no longer needed

PR #149, opened 2025-11-02
feat: add blur node

### LINKED ISSUE #128: Single pixel buffers aren't blurred
I'm not sure if we should include rounded corners as the single pixel buffer is a optimization. We'll have to investigate

### LINKED ISSUE #148: create blur_node
As title suggests. This will greatly simplify our logic while also being more friendly for compositors to use!

the-eater:
This feat adds a blur node that will blur any region underneath it, and handle damage tracking etc (because it already existed, and now I get to use it for free stolen from the other nodes)

### Changes

- Add blur node
- Remove all blur logic from buffer and rect nodes
- Add clipping logic to both blur and texture pass

### Fixes

- Fixes #148 by adding the blur node
- Fixes #128 by decoupling blur from buffer/rect

## Credit

@ErikReider once again for the `struct linked_node`


include/scenefx/types/wlr_scene.h
```diff
@@ -60,6 +60,7 @@ enum wlr_scene_node_type {
 	WLR_SCENE_NODE_SHADOW,
 	WLR_SCENE_NODE_BUFFER,
 	WLR_SCENE_NODE_OPTIMIZED_BLUR,
+	WLR_SCENE_NODE_BLUR_SOURCE,
```
WillPower3309: since this node is now more of a blur node than a blur_source node, I think we can rename it, in this case the enum should be `WLR_SCENE_NODE_BLUR`

include/scenefx/types/wlr_scene.h
WillPower3309: this can be removed

WillPower3309: this can also be removed

types/scene/wlr_scene.c
```diff
@@ -80,6 +73,13 @@ struct wlr_scene_shadow *wlr_scene_shadow_from_node(struct wlr_scene_node *node)
 	return shadow;
 }
 
+struct wlr_scene_blur_source *wlr_scene_blur_source_from_node(struct wlr_scene_node *node) {
+	assert(node->type == WLR_SCENE_NODE_BLUR_SOURCE);
+	struct wlr_scene_blur_source *blur_source = wl_container_of(node, blur_source, node);
+	return blur_source;
+}
+
```
WillPower3309, line 81: nit: remove one of the two empty lines

types/scene/wlr_scene.c
```diff
@@ -2134,6 +2053,34 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
...
+				.corners = blur_source->corners,
+				.discard_transparent = false,
+			},
+			.opaque_region = NULL,
+			.use_optimized_blur = blur_source->should_only_blur_bottom_layer,
+			.blur_data = &scene->blur_data,
+			.ignore_transparent = false,
+			.blur_strength = blur_source->strength,
+		};
+		// Render the actual blur behind the surface
```
WillPower3309: technically this line is no longer needed

types/scene/wlr_scene.c
```diff
@@ -2134,6 +2053,34 @@ static void scene_entry_render(struct render_list_entry *entry, const struct ren
 			});
 		}
 		break;
+	case WLR_SCENE_NODE_BLUR_SOURCE:;
```
WillPower3309: I think we miss using the `transparency_mask_source` here, but I can add that bit to this branch if it's being a pain

the-eater: Added!

types/scene/linked_node.c
```diff
@@ -0,0 +1,87 @@
...
+    return node_1->link != NULL && node_2->link != NULL
+        && node_1->link == node_2->link;
+}
+
+void linked_node_init_link(struct linked_node *node_1,
+        struct linked_node *node_2) {
+    if (node_1->link || node_2->link) {
+        assert(linked_nodes_are_linked(node_1, node_2));
+        linked_node_unlink(node_1, node_2);
+    }
```
WillPower3309, line 42: why are we asserting they're linked only to unlink and then relink them back to eachother?

WillPower3309: we also assert in the unlink method as well

include/scenefx/render/pass.h
```diff
@@ -52,12 +52,23 @@ struct fx_gradient {
...
+};
+
 struct fx_render_texture_options {
 	struct wlr_render_texture_options base;
 	const struct wlr_box *clip_box; // Used to clip csd. Ignored if NULL
 	enum corner_location corners;
 	int corner_radius;
 	bool discard_transparent;
+	struct clipped_region clipped_region;
+	struct fx_render_texture_mask_options mask;
```
WillPower3309: shouldn't this be part of the `blur_pass_options`?

render/fx_renderer/fx_pass.c
```diff
@@ -965,17 +1066,6 @@ void fx_render_pass_add_blur(struct fx_gles_render_pass *pass,
...
-	// Get a stencil of the window ignoring transparent regions
-	if (fx_options->ignore_transparent && fx_options->tex_options.base.texture) {
-		stencil_mask_init();
-
-		struct fx_render_texture_options tex_options = fx_options->tex_options;
-		tex_options.discard_transparent = true;
-		fx_render_pass_add_texture(pass, &tex_options);
-
-		stencil_mask_close(true);
-	}
```
WillPower3309, line 977: I like what you're thinking with getting rid of stenciling, but perhaps this would best be done in a follow-up PR. This would mark the removal of our last stencil usage, so would also involve cleaning some code in https://github.com/wlrfx/scenefx/blob/main/render/fx_renderer/fx_framebuffer.c

We can really reduce the complexity here by using the existing stenciling logic (maybe moved to a new shader as in #133 ), and would avoid the need for the new shaders included in this PR, which we can always add as a follow-up.

render/fx_renderer/fx_pass.c
```diff
@@ -153,40 +155,6 @@ static const struct wlr_render_pass_impl render_pass_impl = {
...
-	glStencilFunc(GL_NOTEQUAL, 1, 0xFF);
-	glStencilOp(GL_KEEP, GL_KEEP, GL_KEEP);
-}
-
-// Finish stenciling and clear the buffer
-static void stencil_mask_fini(void) {
-	glClearStencil(0);
-	glClear(GL_STENCIL_BUFFER_BIT);
-	glDisable(GL_STENCIL_TEST);
-}
```
WillPower3309, line 188: awesome that you worked on this, but as mentioned later maybe best suited for another PR

WillPower3309 (changes requested):
Worth noting as well that the mask bit is irrelevant for RGBX, since we don't have opacity for those nodes (other than the opacity we pass, which would either make it translucent or completely transparent, so no need for masking)

types/scene/linked_node.c
the-eater: pointing at @ErikReider lol

I think the idea behind it was that `init_link` and `unlink` should only ever be called on links that are either both unlinked, or linked to eachother. This one can use an early return rather than an unlink tho

the-eater: I've now turned it into an early return, keeping what I assumed was the assertion, "don't link or unlink 2 unrelated nodes that are linked"

Since I do think that's a fair assertion to make

PR #156, opened 2025-11-24
feat: remove `enum corner_location`

### LINKED ISSUE #137: error when `CORNER_LOCATION_NONE` and `corner_radius` != 0
https://github.com/wlrfx/scenefx/blob/0466cac4be1333d0794b01f4726780be23535767/types/scene/wlr_scene.c#L963-L966

Currently it logs an error if location is none and `corner_radius` != 0. 

But how much of an issue is this? would it make sense to set `corner_radius` to 0 in this function instead of logging of an error?

I am mostly hitting this issue in https://github.com/WillPower3309/swayfx/pull/456

where I use a bitmask to only enable the corners that are applicable. and repeating `(value & mask) != CORNER_LOCATION_NONE ? corner_radius : 0` feels like a bit much

the-eater:
This MR removed the `enum corner_location` and apparently also fixes some issues with scaling.

with SwayFX the changes live look like this:

Settings | Before | After
--- | --- | ---
`transform normal` | 20251125_16h37m06s_grim | 20251125_16h22m42s_grim
`transform 90` | 20251125_16h37m14s_grim | 20251125_16h23m13s_grim
`transform flipped-90` | 20251125_16h37m37s_grim | 20251125_16h23m21s_grim

Closes #137

### DIFF
--- a/include/render/fx_renderer/shaders.h
+++ b/include/render/fx_renderer/shaders.h
@@ -3,7 +3,8 @@
 
 #include <GLES2/gl2.h>
 #include <stdbool.h>
-#include "scenefx/types/fx/clipped_region.h"
+#include <scenefx/types/fx/clipped_region.h>
+#include "types/fx/clipped_region.h"
 
 struct fx_renderer;
 
--- a/include/scenefx/render/pass.h
+++ b/include/scenefx/render/pass.h
@@ -8,6 +8,7 @@
 
 #include "render/egl.h"
 #include "scenefx/types/fx/clipped_region.h"
+#include "types/fx/clipped_region.h"
 
 struct fx_gles_render_pass {
 	struct wlr_render_pass base;
--- a/include/scenefx/types/fx/clipped_region.h
+++ b/include/scenefx/types/fx/clipped_region.h
@@ -1,5 +1,5 @@
-#ifndef TYPES_FX_CLIPPED_REGION_H
-#define TYPES_FX_CLIPPED_REGION_H
+#ifndef SCENEFX_TYPES_FX_CLIPPED_REGION_H
+#define SCENEFX_TYPES_FX_CLIPPED_REGION_H
 
 #include <wlr/util/box.h>
 
@@ -11,13 +11,6 @@ struct fx_corner_radii {
 	uint16_t top_right;
 	uint16_t bottom_right;
 	uint16_t bottom_left;
-};
-
-struct fx_corner_fradii {
-	float top_left;
-	float top_right;
-	float bottom_right;
-	float bottom_left;
 };
 
 #define CORNER_RADIUS_MAX (UINT16_MAX)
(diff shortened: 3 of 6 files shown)


include/render/fx_renderer/shaders.h
WillPower3309, line 1: Seems like these changes actually add some LOC over the original implementation overall, I think we should scrap this bit to keep the diff lean

render/fx_renderer/shaders.c
WillPower3309, line 1: Same as with `shaders.h`

include/scenefx/render/pass.h
```diff
@@ -104,9 +100,8 @@ struct fx_render_blur_pass_options {
 	bool use_optimized_blur;
 	bool ignore_transparent;
 	float blur_strength;
-	int corner_radius;
-	enum corner_location corners;
-	struct clipped_region clipped_region;
+	struct fx_corner_radii corners;
```
WillPower3309: all the other options have a `fradii` type, why not blur_pass_options?

render/fx_renderer/fx_pass.c
```diff
@@ -277,6 +274,30 @@ static void setup_blending(enum wlr_render_blend_mode mode) {
...
+			fmax(clipped_region_box->width - (left + right) * 0.3, 0),
+			fmax(clipped_region_box->height - (top + bottom) * 0.3, 0)
+		);
+		pixman_region32_subtract(clip_region, clip_region, &user_clip_region);
+		pixman_region32_fini(&user_clip_region);
+		return true;
+	}
+
+	return false;
+}
```
WillPower3309, line 299: nice!

include/render/fx_renderer/shaders.h
WillPower3309: I know we save some LOC in fx_pass.c but I think we add more in shaders.c for a positive overall LOC diff, esp considering the `glUniform1f` calls in fx_pass can be oneliners now

include/scenefx/types/fx/clipped_region.h
```diff
@@ -3,12 +3,119 @@
...
+	uint16_t bottom_right;
+	uint16_t bottom_left;
+};
+
+struct fx_corner_fradii {
+	float top_left;
+	float top_right;
+	float bottom_right;
+	float bottom_left;
+};
```
WillPower3309: The fradii stuff is just needed on the scenefx side as far as I could tell, we should move it and its functions to a private header (unless I missed a case)

include/render/fx_renderer/shaders.h
the-eater: I think this is as far as I'm willing to compromise on it, the little LoC added by `shaders.c` is already won back by `shaders.h`

and I am honestly more a DRY person rather than reducing LoC

include/scenefx/types/fx/clipped_region.h
the-eater: fixed

include/scenefx/render/pass.h
the-eater: forgot!

include/render/fx_renderer/shaders.h
WillPower3309: My primary concern was with the shaders.c changes, this is in a fine state to me

PR #158, opened 2025-11-30
feat: implement missing corner radius opacity processing

the-eater:
This code stops nodes with corner radii from being magically transparent, and only makes the affected corners transparent instead

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -294,6 +294,32 @@ static bool scene_nodes_in_box(struct wlr_scene_node *node, struct wlr_box *box,
 	return _scene_nodes_in_box(node, box, iterator, user_data, x, y);
 }
 
+static pixman_region32_t create_corner_location_region(struct fx_corner_radii corners, int x, int y, int width, int height) {
+	pixman_region32_t corner_region;
+	pixman_region32_init(&corner_region);
+	if (corners.top_left) {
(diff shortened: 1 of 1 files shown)


types/scene/wlr_scene.c
```diff
@@ -294,33 +294,57 @@ static bool scene_nodes_in_box(struct wlr_scene_node *node, struct wlr_box *box,
...
+		pixman_region32_init_rect(opaque, x, y, width, height);
+		if (scene_rect->corner_radius > 0 && scene_rect->corners != CORNER_LOCATION_NONE) {
+			pixman_region32_t corners = create_corner_location_region(scene_rect->corner_radius, scene_rect->corners,
+				x, y, width, height);
+			pixman_region32_subtract(opaque, opaque, &corners);
+			pixman_region32_fini(&corners);
+		}
 
+		if (!wlr_box_empty(&scene_rect->clipped_region.area)) {
 			// Subtract the clipped region from a otherwise fully opaque rect
```
WillPower3309: this comment is technically wrong now, we should remove it.
```suggestion
```

types/scene/wlr_scene.c
WillPower3309: I *think* this is now dead code

WillPower3309 (commented):
two small comments

WillPower3309:
Needs a rebase! LGTM other than my one comment

types/scene/wlr_scene.c
the-eater: used an `unreachable();` statement so it hard errors and the compiler is wiser

PR #173, opened 2026-02-23
Renderer and wlr_scene simplifications to match upstream

ErikReider:
This simplifies the logic and "matches" upstream in how things are done which should make rebasing a bit easier for us.

These changes are a bit tangled together, so let me know if it's too difficult to read. Sorry in beforehand @WillPower3309 

Housekeeping:
- Renamed `fx_effect_buffers` -> `fx_offscreen_buffers` which makes more sense
- Moved a lot of logic in wlr_scene to try to ensure clean rebases in the future
- Removed the reliance of a `wlr_output` when drawing blur which should make it more portable

Additions/fixes:
- A more robust blur artifact prevention system which could allow us to add per-blur node `blur_data`
  - Also fixes issues where a blur node wouldn't update when positioned right next to a damaged non-blurred node (like opaque Firefox next to a blurred waybar)
- Be able to choose if offscreen buffer should support alpha or not instead of using the output swapchains drm format
- Make sure to free the offscreen buffers of a `wlr_output` when destroyed
- Better error handling when creating our offscreen buffers
- Fixed potentially binding to an invalid `fbo`, i.e., create one if it is invalid


render/fx_renderer/fx_framebuffer.c
```diff
@@ -1,3 +1,5 @@
+#include <assert.h>
```
WillPower3309: doesn't look like we make any assertions in this PR?

render/fx_renderer/fx_framebuffer.c
```diff
@@ -106,11 +88,37 @@ void fx_framebuffer_get_or_create_custom(struct fx_renderer *renderer,
 		} else {
 			fx_framebuffer_destroy(*fx_framebuffer);
 		}
+		*fx_framebuffer = NULL;
+	}
+
+	// Create a new buffer
+
```
WillPower3309: nit: I think we can remove this comment
```suggestion
```

render/fx_renderer/fx_framebuffer.c
```diff
@@ -1,3 +1,5 @@
+#include <assert.h>
+#include <drm_fourcc.h>
```
WillPower3309: do we use this? It looks like we only use wlr structs to interact with drm

WillPower3309 (commented):
still reviewing but some early comments

render/fx_renderer/fx_pass.c
```diff
@@ -58,14 +101,14 @@ static struct wlr_box get_monitor_box(struct wlr_output *output) {
 
 static const struct wlr_render_pass_impl render_pass_impl;
 
-static struct fx_gles_render_pass *get_render_pass(struct wlr_render_pass *wlr_pass) {
-	assert(wlr_pass->impl == &render_pass_impl);
-	struct fx_gles_render_pass *pass = wl_container_of(wlr_pass, pass, base);
+struct fx_gles_render_pass *fx_get_render_pass(struct wlr_render_pass *render_pass) {
```
WillPower3309, line 104: Q: why is this no longer static? It should remain only visible to this file

ErikReider: We need to get `struct fx_gles_render_pass` in wlr_scene due to the switch to `wlr_renderer_begin_buffer_pass` from `fx_renderer_begin_buffer_pass`

types/scene/wlr_scene.c
```diff
@@ -2975,77 +2910,161 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
...
-			fx_renderer_read_to_buffer(render_pass, &effect_fbos->blur_padding_region,
-					effect_fbos->blur_saved_pixels_buffer,
-					render_pass->buffer);
+	struct fx_gles_render_pass *fx_pass = fx_get_render_pass(render_pass);
+	bool should_compensate_blur = false;
+	if (fx_render_pass_init_offscreen_buffers(render_pass, output)
+			&& pixman_region32_not_empty(&render_data.damage)) {
+		// Blur artifact prevention
+		// Note: Supports individual blur node blur_data
+
```
WillPower3309:
```suggestion
```

types/scene/wlr_scene.c
```diff
@@ -2975,77 +2910,161 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
...
+			}
+
+			// End early. No need to compensate for blur if the whole output
+			// is damaged anyway, but still make sure to let the renderer know that
+			// there's blur nodes to render.
+			if (full_damage) {
+				should_compensate_blur = false;
+				fx_pass->has_blur = true;
+				break;
+			}
```
WillPower3309, line 3054: this check can be done before the loop as we compute `full_damage` much earlier

types/scene/wlr_scene.c
```diff
@@ -2975,77 +2910,161 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
...
+				wlr_region_expand(&intersection, &intersection, sample_size);
+				// Don't re-add already added rectangles
+				pixman_region32_subtract(&intersection, &intersection, &blur_padding_region);
+				pixman_region32_union(&blur_padding_region, &blur_padding_region, &intersection);
+			}
+
+			pixman_region32_fini(&intersection);
+			pixman_region32_fini(&expanded_damage);
+			pixman_region32_fini(&node_visible_region);
 		}
```
WillPower3309: this can be moved into a function for readability

types/scene/wlr_scene.c
```diff
@@ -2975,77 +2910,161 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
...
+				if (!optimized_node->dirty) {
+					continue;
+				}
+			} else {
+				continue;
+			}
+
+			if (!is_scene_blur_enabled(&blur_data)) {
+				continue;
+			}
```
WillPower3309: seeing a lot of `continue`s here. Hopefully with moving the below logic into a function (see my next comment) we an clean some of them up

types/scene/wlr_scene.c
```diff
@@ -2975,77 +2910,161 @@ bool wlr_scene_output_build_state(struct wlr_scene_output *scene_output,
...
+					&blur_padding_region, &render_data.damage);
+			// Make sure that the padding and damage doesn't exceed the output bounds
+			pixman_region32_intersect_rect(&fx_pass->blur_padding_region, &fx_pass->blur_padding_region,
+					0, 0, output->width, output->height);
+
+			// Combine with the render damage (we need to redraw the padding area as well)
+			pixman_region32_union(&render_data.damage,
+					&render_data.damage, &fx_pass->blur_padding_region);
+			// pixman_region32_union(&state->damage,
+			// 		&state->damage, &fx_pass->blur_padding_region);
```
WillPower3309:
```suggestion
```

render/fx_renderer/fx_pass.c
WillPower3309: ah, missed that it was called there

WillPower3309 (changes requested):

types/scene/wlr_scene.c
ErikReider: We need this for letting the optimized_blur_node know that there's blur nodes that need rendering. It's mostly for making the optimized_blur_node wait to render the blur if there are no blur nodes visible.

Should we remove that check?
https://github.com/wlrfx/scenefx/blob/6308c99f6688410ce55978e1cc37a6affecd406e/types/scene/wlr_scene.c#L1960-L1963

WillPower3309 (approved):

PR #175, opened 2026-03-02
add linux/stddef.h for musl compatibility

qaqland:
The __always_inline macro is defined in <linux/stddef.h>. While glibc includes this by default, musl does not. This commit explicitly includes <linux/stddef.h> to ensure compatibility with musl libc platforms.

```
In file included from ../tinywl/tinywl.c:8:
../include/scenefx/types/fx/clipped_region.h:18:8: error: unknown type name '__always_inline'
   18 | static __always_inline uint16_t corner_radius_clamp(int radius) {
      |        ^~~~~~~~~~~~~~~
```

### DIFF
--- a/include/scenefx/types/fx/clipped_region.h
+++ b/include/scenefx/types/fx/clipped_region.h
@@ -1,6 +1,7 @@
 #ifndef SCENEFX_TYPES_FX_CLIPPED_REGION_H
 #define SCENEFX_TYPES_FX_CLIPPED_REGION_H
 
+#include <linux/stddef.h>
 #include <wlr/util/box.h>
 
 // uint16_t is a reasonable enough range for corner radius


ErikReider:
Testing in an alpine:edge docker container, I'm also getting `use of undeclared identifier 'unreachable'`. Are you also experiencing this?

denyallow24:
I mentioned a similar thing in #195 and wrote a patch that defines the macro separately, but I find it more sensible to include a dedicated header. Waiting for any changes.

qaqland:
> Testing in an alpine:edge docker container, I'm also getting `use of undeclared identifier 'unreachable'`. Are you also experiencing this?

Since `unreachable()` was introduced in https://github.com/wlrfx/scenefx/commit/b8a41b6 and later removed in https://github.com/wlrfx/scenefx/commit/235ea6d, this PR should now work correctly.

YsfOuz:
@ErikReider

WillPower3309 (approved):

WillPower3309:
Thanks

PR #177, opened 2026-03-11
fix gles border on nvidia proprietary driver by forcing high prec

apewo:
On nvidia proprietary driver, the border are badly rendered due to medium precision float,

however, forcing high precision float in the fragment shaders and setting the EGL_CONTEXT_CLIENT_VERSION to 3 seems to render the border normally (  tested on : NVIDIA GeForce RTX 2080  /  Driver Version: 590.48.01).

To not  break compatibility with other gpu, this commit add an option : force_fragment_shader_precision_high , so the default build remain unchanged.

to test :  ```meson setup -Dscenefx:force_fragment_shader_precision_high=true ...```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -169,6 +169,10 @@ if get_option('tracy_enable')
 	)
 endif
 
+if get_option('force_fragment_shader_precision_high')
+  add_project_arguments('-DFORCE_GL_FRAGMENT_PRECISION_HIGH=1', language: 'c')
+endif
+
 subdir('protocol')
 subdir('render')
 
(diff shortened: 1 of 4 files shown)


AhjinYeri:
Tested this out myself and it's held up well.

Hardware: RTX 3070 Ti, Nvidia proprietary driver 610.43.03. mango 0.15.5/0.15.6, scenefx 0.5, DankMaterialShell.

I'd been chasing a persistent right-edge border thinning/flickering bug for a while — thinner than expected baseline, and occasional flicker even at idle. Tried a bunch of things on my end without luck. Found this PR linked in MangoWM's Discord bug-reports channel and it seemed relevant.

I built scenefx with this patch applied and swapped it in via LD_LIBRARY_PATH to test without touching my system install. Ran it for about 2.5 hours of normal active use plus idle time, and the issue hasn't recurred once — same machine, same monitors, same config that reliably showed it before.

This would be great to see merged, and maybe eventually on by default for Nvidia proprietary users given how consistently it seems to affect this driver.

apewo:
there is a massive rework of scenefx in the work that amongst other things will include https://github.com/wlrfx/scenefx/pull/179, that patch should fix it and this pr will then be obsolete.

PR #181, opened 2026-04-05
Fixed offscreen_buffers not dropping wlr_buffers

ErikReider:
Should fix the issue that @DreamMaoMao experienced where we were freeing our offscreen GLES2 buffers without dropping the underlying wlr_buffers, leading to a nasty VRAM memory leak when hot-plugging outputs or switching TTYs.


WillPower3309 (approved):
Thanks!

PR #182, opened 2026-04-07
tinywl: Add xdg-shell server as a dependency.

zeckma-tech:
Under some configurations when `-D examples=true` is passed to `meson`, ninja will fail to build the `tinywl/tinywl.p/tinywl.c.o` target because `protocols_server_header['xdg-shell']` wasn't added as a source. `tinywl/tinywl.p/tinywl.c` includes `xdg-shell-protocol.h` which is generated from `protocols_server_header['xdg-shell']`, but if parallelism is too high, or when using samurai, since the dependency information is incomplete, the protocol server header won't generate before `tinywl/tinywl.p/tinywl.c.o` is attempted to be built, and will then proceed to fail.

This commit/PR fixes the issue for all configurations.

### DIFF
--- a/tinywl/meson.build
+++ b/tinywl/meson.build
@@ -1,7 +1,5 @@
 executable(
 	'tinywl',
-	['tinywl.c',
-        protocols_client_header['xdg-shell'],
-        protocols_server_header['xdg-shell']],
+	['tinywl.c', protocols_server_header['xdg-shell']],
 	dependencies: scenefx,
 )
--- a/tinywl/meson.build
+++ b/tinywl/meson.build
@@ -1,5 +1,7 @@
 executable(
 	'tinywl',
-	['tinywl.c', protocols_client_header['xdg-shell']],
+	['tinywl.c',
+        protocols_client_header['xdg-shell'],
+        protocols_server_header['xdg-shell']],
 	dependencies: scenefx,
 )


WillPower3309:
Thanks for the PR! it looks good, but [since upstream no longer contains those lines](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/bbd9a49bdf1449886b8e8e12a96455a1e2228143) I think it's best to follow their path and remove them too.

zeckma-tech:
The server header will still be necessary, unless changed with wlroots-0.20.0. However, the client header shouldn't be necessary anymore, even with 0.19.x. So a simple replace with `client` to `server` will suffice here.

zeckma-tech:
I double checked with wlroots-0.20. Specifically, `/usr/include/wlroots-0.19/wlr/types/wlr_xdg_shell.h` includes `xdg-shell-protocol.h` which is expected to be generated by each build system individually. However, Wayland-Protocols now provides enum headers for most of the protocols. In the case of xdg-shell, this would be `xdg-shell-enum.h`. wlroots-0.20.x's `wlr_xdg_shell.h` now includes the enum header, which doesn't need to be generated by build systems.

So for now, I would recommend merging anyway, but this should be changed when wlroots-0.20 compatibility comes in.

zeckma-tech:
#184 fixes this issue for wlroots-0.20. Considering this superceded by that PR.

PR #183, opened 2026-04-13
WIP: bump scenefx to 0.5, switch to wlroots 0.20

fishman:
For now wlr_scene crashes when moving to 0.20 and I'm not sure 

```gdb
00:00:00.265 [tinywl/tinywl.c:1249] Running Wayland compositor on WAYLAND_DISPLAY=wayland-1
00:00:00.266 [types/output/render.c:123] Attaching empty buffer to output for modeset
00:00:00.266 [types/output/swapchain.c:27] Choosing primary buffer format XR24 (0x34325258) for output 'WL-1'
00:00:00.266 [types/output/swapchain.c:96] Testing swapchain for output 'WL-1'
00:00:00.266 [render/swapchain.c:103] Allocating new swapchain buffer
00:00:00.267 [render/allocator/gbm.c:118] Allocated 697x746 GBM buffer with format XR24 (0x34325258), modifier GFX11,64KB_R_X,PIPE_XOR_BITS=2,PACKERS=2,DCC,DCC_MAX_COMPRESSED_BLOCK=128B,DCC_INDEPENDENT_128B,DCC_PIPE_ALIGN (0x020000001046BB04)
00:00:00.267 [render/fx_renderer/fx_framebuffer.c:153] Created GL FBO for buffer 697x746
00:00:00.270 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
00:00:00.301 [render/swapchain.c:103] Allocating new swapchain buffer
00:00:00.301 [render/allocator/gbm.c:118] Allocated 697x746 GBM buffer with format XR24 (0x34325258), modifier GFX11,64KB_R_X,PIPE_XOR_BITS=2,PACKERS=2,DCC,DCC_MAX_COMPRESSED_BLOCK=128B,DCC_INDEPENDENT_128B,DCC_PIPE_ALIGN (0x020000001046BB04)
00:00:00.302 [render/fx_renderer/fx_framebuffer.c:153] Created GL FBO for buffer 697x746
00:00:00.302 [render/allocator/gbm.c:118] Allocated 697x746 GBM buffer with format XB24 (0x34324258), modifier GFX11,64KB_R_X,PIPE_XOR_BITS=2,PACKERS=2,DCC,DCC_MAX_COMPRESSED_BLOCK=128B,DCC_INDEPENDENT_128B,DCC_PIPE_ALIGN (0x020000001046BB04)
[...]

PR #184, opened 2026-04-24
wlroots-0.20 and bump scenefx to 0.5

### LINKED ISSUE #180: wlroots-0.20 compatibility
wlroots 0.20.0 got released.

List of all breaking changes: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3977

ErikReider:
Supersedes #183 

Todo:
- [ ] ~Fix the fragment shader validation~

### DIFF
--- a/include/types/wlr_scene.h
+++ b/include/types/wlr_scene.h
@@ -5,6 +5,8 @@
 
 struct wlr_scene *scene_node_get_root(struct wlr_scene_node *node);
 
+void scene_node_get_size(struct wlr_scene_node *node, int *width, int *height);
+
 void scene_surface_set_clip(struct wlr_scene_surface *surface, struct wlr_box *clip);
 
 #endif
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
 	'scenefx',
 	'c',
-	version: '0.4.1',
+	version: '0.5.0',
 	license: 'MIT',
 	meson_version: '>=1.3',
 	default_options: [
@@ -66,8 +66,17 @@ else
 	)
 endif
 
+features = {
+	'gles2-renderer': false,
+	'color-management': false,
+}
+internal_features = {
+	'egl': false,
+	'tracy': false,
+}
+
 wayland_kwargs = {
-	'version': '>=1.23.1',
+	'version': '>=1.24.0',
 	'fallback': 'wayland',
 	'default_options': [
 		'tests=false',
@@ -79,8 +88,8 @@ wayland_server = dependency('wayland-server',
 )
 
 wlroots_options = [ 'examples=false' ]
-wlroots_version = ['>=0.19.0', '<0.20.0']
-wlroots = dependency('wlroots-0.19',
+wlroots_version = ['>=0.20.0', '<0.21.0']
+wlroots = dependency('wlroots-0.20',
 	version: wlroots_version,
 	default_options: wlroots_options,
 	required: false,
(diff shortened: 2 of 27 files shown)


fishman:
@ErikReider is there anything missing here? What's wrong with the fragment shaders except for the types?

ErikReider:
> @ErikReider is there anything missing here? What's wrong with the fragment shaders except for the types?

Most shaders don't pass the build-time shader validation due to partial shader files like `gradient.frag` and [this](https://github.com/wlrfx/scenefx/blob/main/render%2Ffx_renderer%2Fshaders%2Ftex.frag#L1-L2). GLES sadly doesn't natively support `#include` pp-macros. We either have to find a way around this or skip shader validation :/

WillPower3309:
I think we can ditch the shader validation - unfortunately this is a limitation of how we currently handle shaders

meson.build
```diff
@@ -66,8 +66,17 @@ else
 	)
 endif
 
+features = {
+	'gles2-renderer': false,
```
WillPower3309: this seems wrong in our case!

fishman:
added it in https://github.com/wlrfx/scenefx/pull/187

WillPower3309 (commented):
some of this appears to be unused by us, but still might be useful as it reduces diff size and will be needed once work begins on the vulkan renderer. Review is still a WIP.

include/types/wlr_scene.h
```diff
@@ -5,6 +5,8 @@
 
 struct wlr_scene *scene_node_get_root(struct wlr_scene_node *node);
 
+void scene_node_get_size(struct wlr_scene_node *node, int *width, int *height);
```
WillPower3309, line 8: weird - this is still only used in `wlr_scene.c`... but I guess whatever upstream decides is best!

render/fx_renderer/fx_pass.c
```diff
@@ -518,39 +519,56 @@ void fx_render_pass_add_rect(struct fx_gles_render_pass *pass,
...
+	const struct wlr_box *clipped_region_box = &fx_options->clipped_region.area;
+	const struct fx_corner_fradii *clipped_region_corners = &fx_options->clipped_region.corners;
+	const bool should_clip = clipped_fregion_is_valid(&fx_options->clipped_region)
+		&& apply_clip_region(&clip_region, clipped_region_box, clipped_region_corners);
+
+	enum wlr_render_blend_mode blend_mode =
+		(color->a == 1.0 && !should_clip) ? WLR_RENDER_BLEND_MODE_NONE : options->blend_mode;
+	const bool use_fast_clear = blend_mode == WLR_RENDER_BLEND_MODE_NONE &&
+			options->clip == NULL &&
+			!should_clip &&
```
WillPower3309: I believe this line is irrelevant

render/fx_renderer/fx_pass.c
```diff
@@ -518,39 +519,56 @@ void fx_render_pass_add_rect(struct fx_gles_render_pass *pass,
...
-	glUseProgram(shader.program);
-	set_proj_matrix(shader.proj, pass->projection_matrix, &box);
-	glUniform4f(shader.color, color->r, color->g, color->b, color->a);
+	push_fx_debug(renderer);
 
-	if (use_clip) {
-		glUniform2f(shader.effects.clip_size, clipped_region_box.width, clipped_region_box.height);
-		glUniform2f(shader.effects.clip_position, clipped_region_box.x, clipped_region_box.y);
-		uniform_corner_radii_set(&shader.effects.clip_radius, &clipped_region_corners);
+	if (use_fast_clear) {
```
WillPower3309, line 532: Q: not sure I'm following the fast clear logic, even if we don't clip and the box matches the buffer, don't we still want to render?

include/types/wlr_scene.h
ErikReider: Apparently only used once in `types/ext_image_capture_source_v1/scene.c` upstream, but yeah, it's mostly there to match upstream :)

render/fx_renderer/fx_pass.c
ErikReider: So instead of rendering a plain, opaque-colored rect through the whole "slow" rendering pipeline, we can tell the GPU to quickly clear the region with a specified color. 

If I remember correctly, this is (on most GPUs) only achievable before the first draw call, so only when we need to clear the whole buffer before rendering a new frame

WillPower3309 (approved):

PR #187, opened 2026-05-21
fragment shader validation

fishman:
- **wlroots-0.20 changes**
- **build(shaders): fix fragment shader validation**

PR #188, opened 2026-05-21
build(shaders): fix fragment shader validation

fishman:
- replace check.sh with python-based preprocessor


WillPower3309:
Am I right in that is #187 but applied to main branch? If so, I'll review and we can close 187

WillPower3309:
Thanks a ton for this though! This looks like a clever solution based on a quick glance

fishman:
Yes that's correct

PR #192, opened 2026-07-06
Vulkan

ErikReider:
Depends on #179 and #190 

Relies on the [wlroots-vkfx](https://gitlab.freedesktop.org/ammen99/wlroots/-/tree/vulkan-effects) fork as there are a few Vulkan functions that aren't publicly exposed by upstream wlroots. Thanks @ammen99! :D

TODO:
- [x] Quad effects
- [ ] Texture effects
- [ ] Blur
  - [ ] Regular blur
  - [ ] Optimized blur
- [ ] Box shadow
- [ ] Quad gradient
- [ ] Quad gradient rounded (merge with regular quad_grad just like the regular quad frag? Could just merge with the regular quad.frag instead?)
- [ ] Blitting for saved blur regions
- [ ] Stenciling
- [ ] Tracy support


WillPower3309:
We can stub the other effects for now and add them incrementally, this is already a lot of great work

WillPower3309:
(I want to be sure I can properly review all the components as well!)

render/vulkan/shaders/corner_alpha.glsl
```diff
@@ -0,0 +1,57 @@
...
+	vec2 bottom_right = abs(relative_pos) - size + radius_br;
+
+	float dist = max(
+		max(get_dist(top_left, radius_tl), get_dist(top_right, radius_tr)),
+		max(get_dist(bottom_left, radius_bl), get_dist(bottom_right, radius_br))
+	);
+
+	float result = smoothstep(0.0, 1.0, dist);
+	return is_cutout ? result : 1.0 - result;
+}
```
WillPower3309, line 57: I think we can have a common shaders dir for glsl bits that are shared between vulkan and gles

fishman:
Oh boy is the wlroots release going to take another year again? 😒

spoloxs:
I have Vulkan blur working on top of this branch — #204. Dual-Kawase plus the padding
repair that keeps damage tracking alive, so `WLR_SCENE_DEBUG_DAMAGE=rerender` is no longer
needed. Blur and HDR compose, since both go through the two-pass pathway.

It needed two additions on the wlroots side, currently living in a fork of ammen99's
`vulkan-effects`:

- Splitting the two-pass render pass into separate scene and output passes, plus
  `wlr_vk_render_pass_suspend()` / `_resume()`, so an effect can end the scene pass
  mid-frame and sample the blend image with neighbourhood taps. An input attachment only
  permits same-pixel reads, and Vulkan forbids ending a pass on a non-final subpass, so
  there was no way to do this from inside the existing pass.
- `SAMPLED` + `TRANSFER_SRC`/`TRANSFER_DST` usage on the blend image, and accessors for its
  view and image, for the snapshot/restore the padding repair performs.

**Before I go further: is there a plan to upstream the renderer-internals export?**
Everything here is gated on it — `wlr_vk_render_pass_get_texture_ds` and friends are not in
upstream `include/wlr/render/vulkan.h`, so my two commits do not apply to wlroots master at
all. I would rather not open an unbuildable MR.

If the wholesale export is a hard sell upstream, would a narrower purpose-built hook be
more likely to land — just the suspend/resume pair and the blend-image accessors, without
exposing descriptor sets and pipelines? Happy to do that work and shape it however is most
useful.

One caveat I should flag on my own change: the split drops `VK_DEPENDENCY_BY_REGION_BIT`, so
subpass merging is lost and the full-screen 16F blend image is stored and reloaded every
frame. Effectively free on desktop discrete GPUs, not free on tilers. Keeping the merged
path when no effect suspends would fix it, and that should probably happen before any of
this goes upstream — especially given
[wlroots#4035](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4035) (Vulkan
renderer efficiency vs GLES2), which this would otherwise make worse.

Possibly useful for framing it upstream:
[wlroots#4043](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4043) already asks
for a blit-based framebuffer API for exactly this — "post-processing pipelines that require
access to intermediate render results", naming blur and bloom. The blend-image accessors
plus `TRANSFER_SRC`/`TRANSFER_DST` are effectively that capability for the Vulkan renderer,
so this may be worth proposing as a partial answer to that issue rather than as a
scenefx-specific export.

PR #193, opened 2026-07-10
vulkan: fix rects with clipped_region rendering as solid quads

m8l8th814n-eng:
Two fixes for `wlr_scene_rect` nodes with a `clipped_region` (the pattern compositors use to draw borders as one rect with the interior clipped out). On the vulkan branch both the fallback path and the vulkan render pass draw such rects as full solid quads, covering everything inside the border.

Symptom as seen in mango (scenefx_init + `WLR_RENDERER=vulkan`, where FX renderer init falls back to base wlroots rendering): the compositor raises the border rect above unfocused clients, so every unfocused window renders as a solid border-colored slab.

**scene: honor rect clipped_region when falling back to base renderer**
When `fx_pass == NULL` (unsupported renderer / fallback mode), `wlr_render_pass_add_rect()` is called with options that have no clipped_region support, so the clipped interior is ignored. Punch the clipped area out of the clip region with pixman before the fallback draw. Corner radii of the clipped region are ignored here (base renderer can't round), which matches the rest of the fallback path.

**vulkan: scissor rect draws with the clipped_region punched out**
`vk_render_pass_add_rect()` computes the punched clip region via `apply_clip_region()` but never uses the result — the draw loop scissors with the original unpunched rects, so the interior is filled and only the shader's corner refinement runs. Intersect the punched region with the pass clip region and scissor with that, matching the gles2 pass.

Tested: first commit verified on a live compositor (mango, vulkan renderer fallback mode) and builds clean with `-Drenderers=gles2` against stock wlroots 0.20. The second commit I could not compile here since wlroots-vkfx isn't public — it only touches the scissor rect source, but please give it a build check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/render/vulkan/pass.c
+++ b/render/vulkan/pass.c
@@ -235,6 +235,11 @@ static void vk_render_pass_add_rect(struct fx_render_pass *fx_pass,
 			pixman_region32_init_rect(&clip_region, box.x, box.y, box.width, box.height);
 		}
 		apply_clip_region(&clip_region, clipped_region_box, clipped_region_corners);
+		pixman_region32_intersect(&clip_region, &clip_region, &clip);
+
+		int scissor_rects_len;
+		const pixman_box32_t *scissor_rects =
+			pixman_region32_rectangles(&clip_region, &scissor_rects_len);
 
 		struct vk_vert_pcr_data vert_pcr_data = {
 			.uv_off = { 0, 0 },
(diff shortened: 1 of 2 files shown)


ErikReider:
Thanks for pointing out that the clipped damage isn't being used when rendering!
There are problems with the solutions though.
1. Falling back to the default rendering path is supposed to fall back to the default behavior. Just cutting out a hole won't work for cases where the hole needs a corner radius as well.
2. This doesn't affect the lines above which marks the original clip region as changed (through `render_pass_mark_box_updated(...)`).

So I'll close this and mention you in the commit as a co-author :)

m8l8th814n-eng:
Appreciate the co-author mention 
Happy to help if I can :)

PR #200, opened 2026-07-27
blur: clip damage compensation to the output, fixing screencast crashes

### LINKED ISSUE #198: Dragging a window between outputs kills screencasts
With blur enabled, dragging a window across the boundary between two outputs instantly kills any screencast of those outputs. The client is disconnected by a protocol error, not just a dropped frame:

```
xdg-desktop-portal-wlr[754568]: ext_image_copy_capture_frame_v1#22: error 2: Invalid buffer damage coordinates
xdg-desktop-portal-wlr[754568]: [ERROR] - wl_display_dispatch failed: Protocol error
systemd[2592]: xdg-desktop-portal-wlr.service: Main process exited, code=exited, status=1/FAILURE
```

PipeWire then drops the node and the consumer ends the stream. Reproduces with both Discord and OBS, so it isn't application-specific.

Also reported on mango (mangowm/mango#1171), which also uses scenefx.

## Reproducing

Two outputs side by side, swayfx config:

```
blur enable
```

1. Start a screencast of one output through xdg-desktop-portal-wlr.
2. Drag any window so it overlaps the boundary between the two outputs
3. The screencast ends immediately as soon as the window hits the overlap and the portal exits

It's a completely consistent crash in either direction (not temperamental at all) regardless of which output the window started on. Disabling blur avoids it.

Instrumenting xdg-desktop-portal-wlr to log the damage it receives shows rectangles arriving outside the captured output, with x tracking the dragged window's left edge past the output origin:

```
-4,282 742x260   (constraints 1920x1080, transform 0)
-8,282 741x260   (constraints 1920x1080, transform 0)
```

## Environment

- scenefx 37ccd723bef49e6891156ffafce8f549f01446cc
- swayfx 0.6 (master [`d3e6784`](/WillPower3309/swayfx/commit/d3e6784b1471e02601055157d3c23a89ce393d19))
- wlroots 0.20.2
- xdg-desktop-portal-wlr 0.8.3
- NVIDIA open kernel module 595.58.03
- Two outputs: 1080x1920 at `0,0` (`transform 270`), 1920x1080 at `1080,420`

ElnuDev:
I encountered https://github.com/wlrfx/scenefx/issues/198 while debugging some issues with my configuration; Claude found this fix and it seems to have fixed the issue on my system and I haven't noticed any regressions so far. Thank you!

<br>

`apply_blur_region()` unions the blur node's visible region, intersected with the expanded frame damage, into both `render_data->damage` and `output_state->damage`.

`node->visible` is in layout coordinates, so a node straddling two outputs reaches outside the output being rendered — negative once translated to output-relative coordinates just above. `render_data->damage` gets clipped to the output later in `scene_output_build_state()`, but `output_state->damage` never is, so those rectangles are published on commit. wlroots forwards them to ext-image-copy-capture sessions, where negative coordinates are a fatal protocol error.

Clip the intersection to the output before publishing it. Damage outside the output can't be rendered into its buffer, so nothing is lost.

The emptiness check is separate: `pixman_region32_intersect()` returns whether the operation succeeded, not whether the result is non-empty, so the old condition was effectively always true. Clipping can now genuinely empty the region, so `should_compensate_blur` needs the real check.

Closes #198

## Testing

Dragging a window across the boundary between my two outputs no longer ends the screencast. Instrumenting xdg-desktop-portal-wlr to log the damage it receives shows nothing arriving outside the output after this change, where before it was every frame of the drag. I have been running it since, with no blur artifacts at the output edges or across the seam.

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -2920,6 +2920,12 @@ static bool apply_blur_region(struct wlr_scene_node *node, struct blur_data *blu
 	pixman_region32_t intersection;
 	pixman_region32_init(&intersection);
 	if (pixman_region32_intersect(&intersection, &expanded_damage, &node_visible_region)) {
+		struct wlr_output *output = render_data->output->output;
+		pixman_region32_intersect_rect(&intersection, &intersection,
+			0, 0, output->width, output->height);
+	}
+
+	if (!pixman_region32_empty(&intersection)) {
 		should_compensate_blur = true;
 
 		// Expand the render damage to re-render surrounding blur nodes


ErikReider (approved):
LGTM, thanks! :D

PR #204, opened 2026-08-10
vulkan: implement blur

spoloxs:
Part of #7. Implements the `Blur → Regular blur` item from #192, on top of that
branch.

Implements blur for the Vulkan renderer, filling in `vk_render_pass_add_blur()`,
`save_blur_region()` and `apply_saved_blur_region()`, which were stubs.

This continues the existing `vulkan-blur` work (shaders, pipelines and the
scenefx-owned sampling path were already in place) and makes blur render.

> [!IMPORTANT]
> **This depends on wlroots changes that are not upstream.** It calls
> `wlr_vk_render_pass_suspend()`, `_resume()`, `_get_blend_image_view()` and
> `_get_blend_image()`, which do not exist in wlroots today. It will not build
> without them. The matching branch is here:
>
> https://github.com/spoloxs/wlroots-vkfx/tree/vulkan-effects
>
> (based on `ammen99/wlroots` `vulkan-effects`, the fork this Vulkan work already
> targets). The code here is complete and reviewable on its own, but it cannot
> build or merge until those wlroots functions have a home — happy to split,
> rebase, or hold this for as long as that takes.

## Why wlroots had to change

Blur has to sample the frame drawn so far. That is impossible from inside the
scene render pass on the two-pass (blending buffer) pathway: geometry and the
blend → output colour transform were two subpasses of a *single* render pass, an
input attachment only permits same-pixel reads (blur needs neighbourhood taps),
and Vulkan forbids ending a render pass while the current subpass is not the
last.

The wlroots branch splits that into two render passes — scene → blend image, then
blend → output carrying the `output_pipe_*` transforms — and exposes
suspend/resume so a consumer can end the scene pass mid-frame, sample the blend
image as a regular texture, and resume with `LOAD_OP_LOAD`.

## How the blur works

Mirrors the GLES2 implementation: N downsample passes (`blur1`), N upsample
passes (`blur2`), then a composite through `blur_effects`.

Render targets live in a new `render/vulkan/effects.c` as plain `VkImage`s owned
by the renderer rather than `wlr_buffer`s, so the chain records into the same
command buffer as the scene and needs no DMA-BUF import or cross-command-buffer
synchronisation.

Their framebuffers are deliberately built against the *scene* render pass:
Vulkan only treats render passes as compatible when they are identical apart
from a short exemption list that does **not** include subpass dependencies, so a
bespoke render pass would make the existing blur pipelines unusable with them.

Because blur only works on the two-pass pathway, it is skipped elsewhere rather
than corrupting the frame. That pathway is also the one colour-managed and HDR
output always take, so blur and HDR compose.

## Notes that cost me time, recorded for the next person

**The composite must use `blur_effects`, not `blur2`.** `blur2` is the upsample
shader and opens with `suv = uv / 2.0`, so using it as a plain blit magnifies the
top-left quarter of the effect image over the whole region — it renders as
ghosted, offset copies of whatever is behind the window.

**The padding repair is what makes damage tracking safe.** Without
`save_blur_region()`/`apply_saved_blur_region()` the blend image still holds the
previous frame's *composited* output wherever nothing was redrawn, so the blur
re-blurs its own result and stair-step banding accumulates. Scoping the chain to
the damage instead just moves the problem: the effect images then have regions
the chain never wrote, which read back as black borders whenever damage is small.

**Copies must be clamped to the image.** `blur_padding_region` is bounded by the
output size, which is not always the render buffer size; a copy running past the
edge is invalid usage and loses the device (`VUID-vkCmdCopyImage-srcOffset-00144`).

`read_to_buffer()` is intentionally left unimplemented here — it is oriented
around `wlr_buffer`s, while the snapshot is a renderer-owned `VkImage`.

## Testing

NVIDIA RTX 5050 (proprietary 610.43.02), Arch, mango 0.15.6, nested Wayland
backend, with `VK_LAYER_KHRONOS_validation` enabled:

- no validation errors across startup, client rendering and teardown
- damage tracking **enabled** (no `WLR_SCENE_DEBUG_DAMAGE=rerender`)
- zero pixel difference between frames captured six seconds apart on an
  otherwise static scene, confirming no blur feedback accumulation
- verified visually against a high-contrast test pattern behind a translucent
  window: smooth blur, correct registration, no black edges or banding

Not yet tested: multi-output, output hotplug/resize under load, non-NVIDIA
drivers, and the `blur_optimized` path (`add_optimized_blur()` is still a stub
returning false).

### DIFF
--- a/include/render/vulkan/pipeline.h
+++ b/include/render/vulkan/pipeline.h
@@ -14,6 +14,9 @@ struct vk_pipeline {
 
 struct vk_pipelines {
 	struct vk_pipeline_quad *quad;
+	struct vk_pipeline_blur *blur1;
+	struct vk_pipeline_blur *blur2;
+	struct vk_pipeline_blur *blur_effects;
 };
 
 void vk_pipelines_init(struct vk_renderer *renderer, struct vk_render_setup *setup,
--- a/include/render/vulkan/shaders.h
+++ b/include/render/vulkan/shaders.h
@@ -66,4 +66,41 @@ struct vk_pipeline *get_vk_quad_pipeline(struct vk_pipeline_quad *quad,
 
 bool vk_shader_info_create_quad(struct vk_renderer *renderer);
 
+///
+/// Blur
+///
+
+// Field order matches the shader: the vec2 comes first so it keeps its 8 byte
+// alignment without needing explicit padding.
+struct vk_frag_blur_pcr_data {
+	float halfpixel[2];
+	float radius;
+};
+
+struct vk_frag_blur_effects_pcr_data {
+	float brightness;
+	float contrast;
+	float saturation;
+	float noise;
+};
+
+struct vk_pipeline_blur {
+	struct vk_pipeline pipeline;
+};
+
+// Shared by blur1/blur2/blur_effects: a single combined image sampler at
+// binding 0. Structurally identical to the layout wlroots uses for its texture
+// descriptor sets, so descriptor sets obtained via
+// wlr_vk_render_pass_get_texture_ds() are layout-compatible with these pipelines.
+VkDescriptorSetLayout vk_get_tex_ds_layout(struct vk_renderer *renderer);
+
+bool vk_shader_info_create_blur1(struct vk_renderer *renderer);
+bool vk_shader_info_create_blur2(struct vk_renderer *renderer);
+bool vk_shader_info_create_blur_effects(struct vk_renderer *renderer);
+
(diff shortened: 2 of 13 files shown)


WillPower3309:
Hey! Just wanted to let you know I'm aware that this exists, thank you for the work and sorry I haven't commented yet! We're trying to focus on the couple of prereq PRs to get in first, namely the renderer refactor and the base vulkan impl.

PR #209, opened 2026-08-28
fix(shadow): handle zero blur sigma

Ly-sec:
## What this fixes

The box-shadow shader evaluates Gaussian functions that divide by the blur
sigma. A sigma of zero therefore produces undefined values and can make a
zero-blur shadow disappear.

Use the existing rounded-corner mask when the sigma is non positive, while
keeping the current Gaussian path for positive values. This also preserves the
configured corner radius for sharp shadows.

## Verification

- `nix build --no-link`
- visual inspection

A visual check with a zero-sigma rounded shadow is still recommended, especially since I'm unsure if this is intended behavior? If so feel free to close it.


WillPower3309 (changes requested):
I like the thinking here, but I want our shaders to be as small as possible, and rendering a rounded rect in this case can instead happen in fx_pass.

Ly-sec:
Hey @WillPower3309 , this should do the trick now.

WillPower3309 (approved):

PR #210, opened 2026-08-28
fix(scene): detect full blur damage by region coverage

Ly-sec:
## What this fixes

A damage region can contain separate rectangles whose combined extents span the
output without actually covering it. The current extents check classifies that
case as full damage and skips blur damage compensation.

Use `pixman_region32_contains_rectangle()` to check whether the damage region
really covers the complete output buffer.

## Verification

- `nix build --no-link`
- Confirmed that two sparse opposite-corner rectangles fail the old full-damage
  predicate and remain partial with the new predicate.
- Confirmed that a rectangle covering the complete buffer is still classified
  as full damage.


DamianDaniel (approved):

WillPower3309:
Thanks a ton!

PR #212, opened 2026-08-31
chore: overall hardening of the repository

DamianDaniel:
### I decided to take a look at the repository and harden it. Here's what i found:
- array_realloc/array_remove_at: fix overflow in doubling and bounds check
- color.c: malloc -> calloc for LUT alloc (overflow check)
- blur setters: clamp num_passes/radius, also clamped in blur_data_calc_size
- optimized_blur_mark_dirty: fix inverted null check
- get_main_buffer_blur: fix dangling pointer (was pointing at a stack local)
- egl.c: fix EGLDisplay leak on two failure paths
- scene-graph.c/tinywl.c: null-check calloc at 6 sites

I have fixed all of these bugs in the hardening branch and am now creating this PR since I consider my work done. It builds clean.

### DIFF
--- a/include/scenefx/types/fx/blur_data.h
+++ b/include/scenefx/types/fx/blur_data.h
@@ -13,6 +13,32 @@ struct blur_data {
 	float saturation;
 };
 
+// Sane, documented bounds for blur_data.num_passes/radius. blur_data_calc_size()
+// feeds pow(2, num_passes + 1) into an int cast, so num_passes needs to stay
+// low enough that the result can't leave int range.
+#define BLUR_DATA_MAX_NUM_PASSES 10
+#define BLUR_DATA_MAX_RADIUS 100
+
+// Clamp num_passes/radius to [0, BLUR_DATA_MAX_*], same pattern as
+// corner_radius_clamp() in clipped_region.h.
+static __always_inline int blur_data_clamp_num_passes(int num_passes) {
+	if (num_passes < 0) {
+		return 0;
+	} else if (num_passes > BLUR_DATA_MAX_NUM_PASSES) {
+		return BLUR_DATA_MAX_NUM_PASSES;
+	}
+	return num_passes;
+}
+
+static __always_inline float blur_data_clamp_radius(float radius) {
+	if (radius < 0) {
+		return 0;
+	} else if (radius > BLUR_DATA_MAX_RADIUS) {
+		return BLUR_DATA_MAX_RADIUS;
+	}
+	return radius;
+}
+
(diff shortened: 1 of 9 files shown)


WillPower3309 (dismissed):
Thanks for this! At a high level, I want to focus these changes on areas that we don't inherit from upstream (in this case that would be the wlr_scene and the fx changes). I also tend to prefer more concise comments. I reviewed the comments in `blur_data.h` to provide an example of what I mean

include/scenefx/types/fx/blur_data.h
```diff
@@ -13,6 +13,32 @@ struct blur_data {
 	float saturation;
 };
 
+// Sane, documented bounds for blur_data.num_passes/radius. blur_data_calc_size()
+// feeds pow(2, num_passes + 1) into an int cast, so num_passes needs to stay
+// low enough that the result can't leave int range.
```
WillPower3309: we can just keep the `num_passes needs to stay low enough that blur_data_calc_size() can't leave int range`

include/scenefx/types/fx/blur_data.h
```diff
@@ -13,6 +13,32 @@ struct blur_data {
...
+#define BLUR_DATA_MAX_NUM_PASSES 10
+#define BLUR_DATA_MAX_RADIUS 100
+
+// Clamp num_passes/radius to [0, BLUR_DATA_MAX_*], same pattern as
+// corner_radius_clamp() in clipped_region.h.
```
WillPower3309: comment not needed, implied by function name

DamianDaniel: fixed in commit f19fe88e2307c2e49df6c6a86208e6bc49d5b346

include/scenefx/types/fx/blur_data.h
DamianDaniel: f19fe88e2307c2e49df6c6a86208e6bc49d5b346

DamianDaniel:
> Thanks for this! At a high level, I want to focus these changes on areas that we don't inherit from upstream (in this case that would be the wlr_scene and the fx changes). I also tend to prefer more concise comments. I reviewed the comments in `blur_data.h` to provide an example of what I mean

i rewrote/deleted the comments and reverted to upstream the way you requested in commit <a href="/wlrfx/scenefx/commit/752726cf80774d35c4a8a3a2879c12c6eeb30c8a">f19fe88e2307c2e49df6c6a86208e6bc49d5b346</a>. I think it's ready to merge now. If not, state the reason on and ill happily change it.

PR #220, opened 2026-09-18
build: add shared library versioning

### LINKED ISSUE #219: build: soname
Currently the built library is named `libscenefx-<version>.so` (e.g.
`libscenefx-0.5.so`), with the version baked directly into the target
name in meson.build, and no `.so.X` soname/symlink chain.

Standard practice for shared libraries on Linux is:
- `libscenefx.so.X.Y.Z` -> actual file
- `libscenefx.so.X` -> soname symlink (X = ABI/soversion)
- `libscenefx.so` -> dev symlink (used only at link time, ships in -devel)

This came up during Fedora packaging review (guidelines mark it as a
"should", not a hard blocker: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shared_libraries).

Fix would be adding `version:`/`soversion:` arguments to the
`library()` call in meson.build, so meson generates the proper
symlink chain and embeds the correct SONAME in the ELF header,
instead of the version being part of the literal target name.

Same likely applies to the pkgconfig file name and the versioned
include directory (`scenefx-<version>.pc`, `/usr/include/scenefx-<version>/`),
worth deciding whether those should also drop the embedded version
or stay as-is (parallel-install scheme) once soname is fixed.

Suhail98:
Add Meson's `version` and `soversion` arguments to the existing library target so SceneFX installs a versioned shared library and SONAME. For 0.5.0:

```text
libscenefx-0.5.so -> libscenefx-0.5.so.0 -> libscenefx-0.5.so.0.5.0
SONAME: libscenefx-0.5.so.0
```

Keep `versioned_name` to preserve the parallel-install scheme from #103. The linker flag, pkg-config name, Meson dependency name, and installed header paths remain unchanged. This follows the existing version variables: the major version supplies the ABI suffix within the major.minor library namespace.

This fixes the missing SONAME versioning and symlink chain while retaining the versioned library basename, rather than switching to the generic name suggested in #219. Downstream binaries need rebuilding to use the new SONAME when the unversioned linker symlink moves into the development package.

Fixes #219

## Testing

Validated commit `516edd6` on x86_64 Linux:

- Full Nix build with the repository's unchanged `flake.lock`.
- Regression check: upstream `714c0de` builds and installs successfully, but fails the ABI-versioned SONAME assertion because its ELF SONAME is `libscenefx-0.5.so` and `.so.0` is absent. The identical assertion passes on this PR's real installed library, with both symlinks verified. This exercises normal Meson build/install output, not a manually substituted runtime handler.
- Full Meson shared, static, and both-library builds and staged installs, including `tinywl` and `scene-graph`. Verified SONAME, symlinks, headers, pkg-config, and example ELF dependencies.
- Compiled and ran an installed-library consumer that creates and destroys a scene/tree.
- Separate Meson 1.3.2 fixtures verified the minimum supported Meson series and side-by-side installation of simulated 0.5/0.6 versions.
- Built Fedora Rawhide RPMs using [Damian's COPR spec](https://download.copr.fedorainfracloud.org/results/damiand/scenefx/srpm-builds/10984315/scenefx-0.5-1.src.rpm) with this commit's source and the file-list adjustment below. The consumer ran after removing `scenefx-devel`, and in a fresh container with only the runtime RPM and automatically resolved dependencies. No development packages, RPATH/RUNPATH, or LD_LIBRARY_PATH were needed in that fresh container; `rpm -V scenefx` passed.
- `git diff --check` passed. The project defines no Meson tests. GPU rendering was not exercised.

## Downstream packaging

The existing COPR spec rejects the new `.so.*` files as unpackaged. The tested adjustment is to move the unversioned linker symlink to `-devel` and keep the ABI symlink and real library in the runtime package:

```spec
%files
# Keep existing license/documentation entries.
%{_libdir}/libscenefx-%{version}.so.*

%files devel
%{_libdir}/libscenefx-%{version}.so
%{_libdir}/pkgconfig/scenefx-%{version}.pc
%{_includedir}/scenefx-%{version}/
```

This spec adjustment was tested locally; nothing was published to COPR.


DamianDaniel (approved):
This is honestly a well-done PR. Although the changes aren't big you went out of your way to test it properly. We'll be waiting for a review from @WillPower3309 and can merge it. 
Thanks for your contribution.

--
Damian

WillPower3309:
Thanks!

